PR #3, opened 2020-04-06
Automate creation of ACM cert and R53 domain

bmonkman:


bmonkman:
Going to add optional creation of a pretty name for CF.

bmonkman:
On second thought I'm not going to make it optional. It doesn't make much sense to leave that out because without a domain pointing at this stuff it's kind of pointless..

PR #6, opened 2020-04-14
rds: db monitoring role to be created by TF

davidcheung:
RDS creation does not guarentee the enhance_monitoring role is created
before TF tries to use the role, creating it will stop it from erroring

### DIFF
--- a/terraform/modules/database/main.tf
+++ b/terraform/modules/database/main.tf
@@ -35,12 +35,12 @@ data "aws_secretsmanager_secret_version" "rds_master_secret" {
 
 module "rds" {
   source  = "terraform-aws-modules/rds/aws"
-  version = "2.4.0"
+  version = "2.14.0"
 
   identifier = "${var.project}-${var.environment}"
 
   engine            = "postgres"
-  engine_version    = "11.5"
+  engine_version    = "11"
   instance_class    = var.instance_class
   allocated_storage = var.storage_gb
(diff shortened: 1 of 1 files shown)


terraform/modules/database/main.tf
```diff
@@ -71,8 +71,10 @@ module "rds" {
 
   # Enhanced monitoring
   performance_insights_enabled = true
+  create_monitoring_role = true
+  monitoring_role_name = "${var.project}-${var.environment}-rds-monitoring-role"
+  monitoring_role_arn = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/${var.project}-${var.environment}-rds-monitoring-role"
```
bmonkman, line 76: The ARN is not required when the previous two parameters are supplied. It will create a role and then use that ARN.

davidcheung: I thought so too, it actually errors out when you don't provide it

bmonkman: What sort of error?

davidcheung: >InvalidParameterCombination: A MonitoringRoleARN value is required if you specify a MonitoringInterval value other than 0.

Feels like some parameter validation enforcement.

bmonkman: Yeah let's bump that version up!

Issue #9, opened 2020-04-17
Add k8s dashboard to the cluster by default


bmonkman:
This has been a request for most of the clusters we have set up, let's include it by default. At some point when we add more parameters we can let people turn it off if they don't want it.

PR #12, opened 2020-04-17
#9: Added kubernetes dashboard and metrics scraper, plus some documen…

bmonkman:
…tation


kubernetes/terraform/modules/kubernetes/kubernetes_dashboard.tf
```diff
@@ -0,0 +1,309 @@
...
+          liveness_probe {
+            http_get {
+              path   = "/"
+              port   = "8443"
+              scheme = "HTTPS"
+            }
+            initial_delay_seconds = 30
+            timeout_seconds       = 30
+          }
+          image_pull_policy = "Always"
```
ShahNewazKhan, line 216: Given the image is pinned to tag `v2.0.0-rc7` we can set the `image_pull_policy` to `IfNotPesent` as it is unnecessary to re-pull the same image.

PR #13, opened 2020-04-20
Forgot to add clusterissuer for cert-manager

bmonkman:
Thanks for the heads up @davidcheung !

### DIFF
--- a/kubernetes/terraform/modules/kubernetes/cert_manager.tf
+++ b/kubernetes/terraform/modules/kubernetes/cert_manager.tf
@@ -1,7 +1,7 @@
 locals {
   cert_manager_namespace   = "kube-system"
   cert_manager_version     = "0.14.2"
-  cluster_issuer_name      = "clusterissuer-letsencrypt-${var.cert_manager_acme_environment}"
+  cluster_issuer_name      = var.cert_manager_use_production_acme_environment ? "clusterissuer-letsencrypt-production" : "clusterissuer-letsencrypt-production"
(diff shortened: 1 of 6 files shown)


kubernetes/terraform/modules/kubernetes/cert_manager.tf
```diff
@@ -1,6 +1,8 @@
 locals {
   cert_manager_namespace  = "kube-system"
   cert_manager_version    = "0.14.2"
+  cluster_issuer_name      = "clusterissuer-letsencrypt-${var.cert_manager_acme_environment}"
+  cert_manager_acme_server = var.cert_manager_acme_environment == "production" ? "https://acme-v02.api.letsencrypt.org/directory" : "https://acme-${var.cert_manager_acme_environment}-v02.api.letsencrypt.org/directory"
```
davidcheung, line 5: should we just hardcode the non-production to `staging`?

trevoro: @bmonkman @davidcheung I've been wondering about this. We could also just remove multiple environments from the results here and have everything be in one environment, rather than forcing this. AFAIK we dont force people to apply staging and production either; the Makefile takes an argument to try it in PROD.

bmonkman: This is not our environment, it's the LetsEncrypt environment, and they aren't 1:1.
For example, you would want to use the staging LE env for testing but then production when your config is ready, and most likely both your staging and production environments want to use the production LE env because that's the only one that gives valid certs.

bmonkman: And yeah David, that's probably fine. I thought they did have other environments but maybe not, and even if they do, that's a bit too deep to go down this rabbit hole.

PR #14, opened 2020-04-20
Merge secret

davidcheung:
secret names are unique and long-lived(soft delete) and
names are not reusable, so for development/shared env purposes it
will be nicer to have secrets use a name_prefix to avoid name conflicts

Note: originally secret was put in a separate TF state so the secrets
can be referenced by name only, but RDS does not support secret-manager

### DIFF
--- a/terraform/modules/environment/main.tf
+++ b/terraform/modules/environment/main.tf
@@ -51,7 +51,7 @@ module "eks" {
 }
 
 data "aws_iam_user" "ci_user" {
-  user_name = "ci-user" # Should have been created in the bootstrap process
+  user_name = "${var.project}-ci-user"  # Should have been created in the bootstrap process
 }
 
 module "wildcard_domain" {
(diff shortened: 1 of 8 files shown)


terraform/modules/environment/main.tf
```diff
@@ -50,8 +50,23 @@ module "eks" {
...
+  name = "${var.project}-ci-user"
+}
+
+# Create a keypair to be used by CI systems
+resource "aws_iam_access_key" "ci_user" {
+  user    = aws_iam_user.ci_user.name
+}
+
+# Add the keys to AWS secrets manager
+module "ci_user_keys" {
```
bmonkman, line 64: Hmm we could have left this in the secrets section. It still might be beneficial to have that for secrets like this that we just want to put into secrets manager and forget about. I know it's another step and that's annoying, but we should be careful how many secrets get put into the TF state.

terraform/modules/secret/variables.tf
```diff
@@ -1,5 +1,5 @@
-variable "name" {
-  description = "The name of the secret in Secrets Manager"
+variable "name_prefix" {
+  description = "The name prefix of the secret in Secrets Manager"
```
bmonkman, line 2: Could you add a default for this? Since it's really only for testing and not something that normal users should care about we should try to keep it under the hood as much as possible.

PR #20, opened 2020-04-24
K8s utils qol

davidcheung:
1. the jetstack cert-manager out of the box only supports cert-manager being on [cert-manager namespace][1]

2. executing the local files requires the kubeconfig to be updated beforehand

3. kube dashbaord sometimes spins up before the rolebinding 😢 

[1]: https://github.com/commitdev/commit0-aws-eks-stack/blob/a054f7b8d6e0da01494256618e65161f185faaab/kubernetes/terraform/modules/kubernetes/files/cert-manager.crds.yaml#L46

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,10 +17,12 @@ apply-env:
 	terraform init; \
 	terraform apply
 
-apply-k8s-utils:
+apply-k8s-utils: update-k8s-conf
 	pushd kubernetes/terraform/environments/$(ENV); \
 	terraform init; \
 	terraform apply
+
+update-k8s-conf: eks --region <% index .Params `region` %> update-kubeconfig --name <% .Name %>-$(ENV)-<% index .Params `region` %>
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -18,10 +18,14 @@ apply-env:
 	terraform apply
 
 apply-k8s-utils:
+	update-k8s-conf; \
```
bmonkman, line 21: This should be on the line above like: `apply-k8s-utils: update-k8s-conf`

Makefile
```diff
@@ -18,10 +18,14 @@ apply-env:
...
 	pushd kubernetes/terraform/environments/$(ENV); \
 	terraform init; \
 	terraform apply
 
+update-k8s-conf:
+	aws eks --region <% index .Params `region` %> update-kubeconfig --name <% .Name %>-$ENV-<% index .Params `region` %>
```
bmonkman, line 27: Should be $(ENV)

PR #23, opened 2020-04-27
Fix the kubernetes admin role, pull some of the iam stuff out into a …

bmonkman:
…separate tf file

### DIFF
--- a/terraform/modules/environment/iam.tf
+++ b/terraform/modules/environment/iam.tf
@@ -1,0 +1,65 @@
+
+# Data sources for EKS IAM
+data "aws_caller_identity" "current" {}
+
+# @TODO - sort out creating only a single user but multiple roles per env
+
+# Create KubernetesAdmin role for aws-iam-authenticator
+resource "aws_iam_role" "kubernetes_admin_role" {
+  name               = "${var.project}-kubernetes-admin-${var.environment}"
+  assume_role_policy = data.aws_iam_policy_document.assumerole_root_policy.json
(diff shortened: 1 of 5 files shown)


terraform/modules/s3_hosting/main.tf
```diff
@@ -126,6 +126,10 @@ resource "aws_cloudfront_distribution" "client_assets_distribution" {
       ssl_support_method       = "sni-only"
     }
 
+  # Reference the cert validations only so it becomes a dependency
+  tags = {
```
davidcheung, line 130: nice, didnt know you could do that

bmonkman: TF builds a big dependency graph so it can decide what can be run in parallel and what needs to be serial. The problem here though is just that it's too smart. I thought just by providing the validations resource as an argument to this module it would create a dependency, but it looks like we actually have to use it inside the module too.

terraform/modules/environment/iam.tf
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }
+
+  # Allow the CI user to assume this role
+  statement {
+    actions = ["sts:AssumeRole"]
+
+    principals {
+      type        = "AWS"
+      identifiers = [data.aws_iam_user.ci_user.arn]
```
davidcheung, line 31: it can access the `ci_user` from the main file right?

bmonkman: Yes, since they are in the same directory. Though it looks like I forgot to remove something from main that I copied over. Thanks!

PR #27, opened 2020-04-28
Change parameter names to fit new format from the doc we are working …

bmonkman:
…off, and so zero.yml can be shared between all projects

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Then get the aws account id:
 aws sts get-caller-identity --query "Account"
 ```
 
-Then create a `config.yml` file and fill in the appropriate values:
+Then create a `zero.yml` file and fill in the appropriate values:
 
 ```yaml
 name: my-project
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -48,8 +48,13 @@ params:
...
   accountId: <from above>
   kubeWorkerAMI: ami-<from above>
-  productionHost: domain.com
-  stagingHost: domain-staging.com
+  productionHostRoot: domain.com
+  productionFrontendHost: domain.com
+  productionBackendHost: api.domain.com
+  stagingHostRoot: domain-staging.com
+  stagingFrontendHost: domain-staging.com
+  stagingBackendHost: api.domain-staging.com
```
trevoro, line 56: More like "lets talk about this later" but this is the kind of place where we'll benefit by having these configuration values but also providing reasonable defaults too.

bmonkman: I think these are already good suggestions. Use different domains for staging and production, have the front end served from the root, provide your backend on a subdomain called `api`

PR #28, opened 2020-04-28
fixing k8s config target

trevoro:
If the Makefile command is not on a new tab indented line then it treats the following word as a target name with arguments. This change just ensures the command is executed. Also added the word `aws` in front as that binary name was missing.


trevoro:
All good!

PR #32, opened 2020-04-29
random keepers: terraform state to keep state

davidcheung:
I'm still testing this, i tried it and couldnt seem to get it to work... 
I'm wondering if its the way `"${data.aws_secretsmanager_secret_version..secret_strirds_master_secretng}"` is read, it always thinks its new (maybe it doesnt keep a state of it?) , or maybe i did something wrong

### DIFF
--- a/terraform/modules/database/main.tf
+++ b/terraform/modules/database/main.tf
@@ -45,7 +45,6 @@ data "aws_secretsmanager_secret" "rds_master_secret" {
 # RDS does not support secret-manager, have to provide the actual string
 data "aws_secretsmanager_secret_version" "rds_master_secret" {
   secret_id = data.aws_secretsmanager_secret.rds_master_secret.name
-  depends_on = [data.aws_secretsmanager_secret.rds_master_secret]
 }
 
 module "rds" {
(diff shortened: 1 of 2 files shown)


bmonkman:
So you have applied that multiple times and it still keeps generating new ones?

terraform/modules/database/main.tf
```diff
@@ -45,7 +45,6 @@ data "aws_secretsmanager_secret" "rds_master_secret" {
 # RDS does not support secret-manager, have to provide the actual string
 data "aws_secretsmanager_secret_version" "rds_master_secret" {
   secret_id = data.aws_secretsmanager_secret.rds_master_secret.name
-  depends_on = [data.aws_secretsmanager_secret.rds_master_secret]
```
davidcheung, line 48: >So you have applied that multiple times and it still keeps generating new ones?

@bmonkman yeah, i have finally been able to pin point the problem, its the depends_on
not sure I understand why though

davidcheung: and i added it trying to solve the initial creation where it errors out saying 
`AWSCURRENT version not found` or something along the lines of that

bmonkman: That shouldn't require `depends_on` because the line above it already creates an explicit dependency. Weird that it would cause (or fix) any problems though...

PR #39, opened 2020-05-04
separate remote-state by env

davidcheung:


terraform/bootstrap/remote-state/production/main.tf
```diff
@@ -0,0 +1,8 @@
+module "production_tf_state" {
+    source = "../../../modules/remote-state"
+    environment = "production"
```
bmonkman, line 3: We already have the environment provided as a variable when apply is run. How about just use that var here so we don't have to duplicate the files?

terraform/bootstrap/remote-state/staging/main.tf
```diff
@@ -0,0 +1,8 @@
+module "staging_tf_state" {
+    source = "../../../modules/remote-state"
+    environment = "staging"
+}
+
+variable "environment" {
+    description = "The environment (development/staging/production)"
```
karimkawambwa, line 7:
```suggestion
    description = "The environment (development/staging/production)"
    default = "staging"
```

terraform/bootstrap/remote-state/production/main.tf
```diff
@@ -0,0 +1,8 @@
...
+}
+
+variable "environment" {
+    description = "The environment (development/staging/production)"
```
karimkawambwa, line 7:
```suggestion
    description = "The environment (development/staging/production)"
    default = "production"
```

terraform/bootstrap/remote-state/production/main.tf
davidcheung: as in not moving it to a module?
I would still need 1 set of the file per env, right? 

or you mean just using `environment = var.environment`

bmonkman: Yeah, just using the variable.

Makefile
```diff
@@ -3,7 +3,7 @@ ENV ?= staging
 apply: apply-remote-state apply-secrets apply-env apply-k8s-utils
 
 apply-remote-state:
-	pushd terraform/bootstrap/remote-state; \
+	pushd terraform/bootstrap/remote-state/$(ENV); \
 	terraform init; \
 	terraform apply -var "environment=$(ENV)"
```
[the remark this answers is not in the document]
davidcheung: decided to go with the `var.environment` approach as Bill suggested so will need to keep that

terraform/bootstrap/remote-state/production/main.tf
davidcheung: i'm going to remove `development` from the text as well 😅 , since if you can reach those terraform file means youre never `env=developement`

terraform/bootstrap/remote-state/production/main.tf
bmonkman: Actually I think even having it as a module is overkill. How about just leave it as it was before, but accept the environment variable?

davidcheung: sure, either way is easy to change down the road

bmonkman:
Now can you just move `main.tf` up so it's not inside an environment directory?

davidcheung:
Not sure if I understand, you mean `/remote-state/production/main.tf` -> `/remote-state/main.tf`? 
this would go back to the original behavior where they end up sharing state across envs no? 
>Now can you just move main.tf up so it's not inside an environment directory?

bmonkman:
This is part of a bigger discussion actually. The local state here is what's causing the problem, and I think the way we are dealing with that is not quite right. We can't rely on the local state being present for this because it won't be checked in to source control. And if that's the case we should probably just remove it so people don't depend on it. Though that does make it a bit harder to test with because we can't have a `teardown` command for the remote state that we bootstrap. 
I think there are 2 options here:

1. Delete the local state file after this operation. If someone wants to tear down the environment they would have to manually delete the s3 bucket and dynamodb table. This is fine for normal use cases, as most likely you will create the remote state and leave it forever. It's a pain for testing when you may need to delete and recreate.

2. Write out different state files with the `-state=production.tfstate` and check them in. Having state in git is not great, and this is not necessary for most cases, but it makes it easer for testing.

I guess a hybrid approach could be to write out state files for the different environments and then exclude them with gitignore, and use them only for testing. But I would be a bit worried that this would be misleading for people.
Or maybe we need some separate make targets just for testing. 
Thoughts?

davidcheung:
the teardown already often requires manually emptying the S3 bucket, we could alias some commands to do what the `terraform destroy` since its only 2 resources to replace the `make teardown` 

if we do option2 with `-state=production.tfstate` <- this means we will treat tfstate as a first class citizen and tie it into our infra? and pull out the correct state based on `ENV=?` 

Hmm, is having 2 folders for 2 env's TF local state that bad? I feel this is the less confusing route.  And all the other stuff would go into the correct remote state since theyre initiated with the terraform backend with `terraform/environments/production/main.tf` etc  (or you're talking about removing all the environment folders altogether?)

bmonkman:
The problem with having directories per environment is that it implies that there should be local state which is not the case. 
I'm kinda leaning toward wanting to just delete the local state files after they are created to reinforce the fact that it shouldn't be relied upon. 
And if we really want to have some helpers for testing maybe we just delete those resources using the aws cli like you mentioned. 
Same goes for secrets.

Makefile
```diff
@@ -2,10 +2,11 @@ ENV ?= staging
...
 
+## remove state file only if exit code 0 from terraform apply
 apply-remote-state:
 	pushd terraform/bootstrap/remote-state; \
 	terraform init; \
-	terraform apply -var "environment=$(ENV)"
+	terraform apply -var "environment=$(ENV)" && rm ./terraform.tfstate;
 
 apply-secrets:
 	pushd terraform/bootstrap/secrets; \
```
davidcheung, line 12: running into some trouble with the teardown command for secrets, since it was changed to a name prefix, ill need to get the nameback to delete it 😢

Makefile
```diff
@@ -2,10 +2,11 @@ ENV ?= staging
...
 
+## remove state file only if exit code 0 from terraform apply
 apply-remote-state:
-	pushd terraform/bootstrap/remote-state/$(ENV); \
+	pushd terraform/bootstrap/remote-state; \
 	terraform init; \
-	terraform apply -var "environment=$(ENV)"
+	terraform apply -var "environment=$(ENV)" && rm ./terraform.tfstate;
 
 apply-secrets:
```
bmonkman, line 11: We should do the same for secrets

Makefile
```diff
@@ -28,12 +29,15 @@ update-k8s-conf:
...
+	export AWS_PAGER=''; \
+	aws s3 rb s3://<% .Name %>-$(ENV)-terraform-state --force; \
+	aws dynamodb delete-table --table-name <% .Name %>-$(ENV)-terraform-state-locks;
 
 teardown-secrets:
-	pushd terraform/bootstrap/secrets; \
-	terraform destroy -auto-approve;
+	export AWS_PAGER=''; \
+	aws secretsmanager  list-secrets --query "SecretList[?Tags[?Key=='project' && Value=='<% .Name %>']].[Name]" | jq  '.[] [0]' | xargs aws secretsmanager delete-secret --secret-id; \
+	aws iam delete-access-key --user-name <% .Name %>-ci-user --access-key-id $(shell aws iam list-access-keys --user-name <% .Name %>-ci-user --query "AccessKeyMetadata[0].AccessKeyId" | sed 's/"//g'); \
```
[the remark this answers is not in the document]
davidcheung: they actually dont cascade delete entities it seems, i get the following error without deleting the keys
 >An error occurred (DeleteConflict) when calling the DeleteUser operation: Cannot delete entity, must delete access keys first.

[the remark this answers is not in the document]
bmonkman: Okay, one more thing to try.. The `--access-key-id` param to `delete-access-key` takes a regex. Could you just say `-access-key-id .*`?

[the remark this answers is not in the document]
davidcheung: their man page is lying i suspect, it may support regex but it needs to go through a bunch of validation that conforms the keyID's schema 🤦 

with `.*?` 
>Parameter validation failed:
Invalid length for parameter AccessKeyId, value: 3, valid range: 16-inf

i tried to make it 16 character or above with `\w\w\w\w\w\w\w\w[\w]+` then it gave me 
>An error occurred (ValidationError) when calling the DeleteAccessKey operation: The specified value for accessKeyId is invalid. It must contain only alphanumeric characters.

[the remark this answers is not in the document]
bmonkman: Sheesh. Oh well, thanks for trying!

PR #45, opened 2020-06-23
Added features necessary to support the new zero apply command

bmonkman:
Added optional auto approve in make targets
Added checking for existing resources and skipping bootstrap steps
Added random seed for secret suffixes
Upgraded EKS version to 1.16 and module to 12.1.0

### DIFF
--- a/templates/terraform/bootstrap/secrets/main.tf
+++ b/templates/terraform/bootstrap/secrets/main.tf
@@ -26,7 +26,7 @@ resource "aws_iam_access_key" "ci_user" {
 module "ci_user_keys" {
   source  = "../../modules/secret"
 
-  name_prefix    = "ci-user-aws-keys"
+  name    = "ci-user-aws-keys<% index .Params `randomSeed` %>"
   type    = "map"
   values  = map("access_key_id", aws_iam_access_key.ci_user.id, "secret_key", aws_iam_access_key.ci_user.secret)
   tags = map("project", local.project)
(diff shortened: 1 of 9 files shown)


zero-module.yml
```diff
@@ -44,4 +44,10 @@ parameters:
   - field: eksWorkerAMI
     label: EKS Worker EC2 AMI ID
     execute: aws ssm get-parameters --names /aws/service/eks/optimized-ami/1.15/amazon-linux-2/recommended/image_id --region $region --query "Parameters[0].Value" | tr -d '"'
+  - field: eksWorkerAMI
+    label: EKS Worker EC2 AMI ID
+    execute: aws ssm get-parameters --names /aws/service/eks/optimized-ami/1.15/amazon-linux-2/recommended/image_id --region $region --query "Parameters[0].Value" | tr -d '"'
```
cjhawkins, line 49: Should this be checking `/aws/service/eks/optimized-ami/1.16/`

templates/terraform/bootstrap/secrets/main.tf
```diff
@@ -26,7 +26,7 @@ resource "aws_iam_access_key" "ci_user" {
 module "ci_user_keys" {
   source  = "../../modules/secret"
 
-  name_prefix    = "ci-user-aws-keys"
+  name    = "ci-user-aws-keys<% index .Params `randomSeed` %>"
```
cjhawkins, line 29: Just a general question... is randomSeed used to create a unique namespace that modules share for secrets, or does it actually get as a seed in random number generation by the modules?

bmonkman: It just gets used as a suffix for the secret name, as secrets have an annoying behaviour where they are not fully deleted, so if you delete a secret and then try to use the same name again it fails. Not a big deal in the real world but it's annoying for testing.

PR #46, opened 2020-06-26
secrets: remove name_prefix fixes

davidcheung:
just realized go-getter supports branches so we can test with a branch doing, or even offer different flavors of our modules via branches 

```
func GetRegistry() Registry {
	return Registry{
		// TODO: better place to store these options as configuration file or any source
		{
			"EKS + Go + React",
			[]string{
				"github.com/commitdev/zero-aws-eks-stack?ref=fix-secret",
				"github.com/commitdev/zero-deployable-backend",
				"github.com/commitdev/zero-deployable-react-frontend",
			},
		},
		{
			"Custom", []string{},
		},
	}
}
```

### DIFF
--- a/templates/terraform/modules/database/main.tf
+++ b/templates/terraform/modules/database/main.tf
@@ -32,8 +32,7 @@ data "aws_caller_identity" "current" {
 module "db_password" {
   source      = "../secret"
   type        = "random"
-  ## note: secret name_prefix has a limitation of 32 characters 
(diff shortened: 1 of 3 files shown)


bmonkman:
> just realized go-getter supports branches so we can test with a branch doing, or even offer different flavors of our modules via branches

You can also point at local directories which is is probably even easier for testing.

PR #47, opened 2020-06-26
fixes for helm_release syntax updates

davidcheung:


templates/Makefile
```diff
@@ -32,16 +32,21 @@ update-k8s-conf:
...
 	aws dynamodb delete-table --table-name <% .Name %>-$(ENVIRONMENT)-terraform-state-locks
 
 teardown-secrets:
-	export AWS_PAGER='' && \
+	export AWS_PAGER='' && export AWS_DEFAULT_REGION=<% index .Params `region` %> && \
 	aws secretsmanager list-secrets --query "SecretList[?Tags[?Key=='project' && Value=='<% .Name %>']].[Name] | [0][0]" | xargs aws secretsmanager delete-secret --secret-id && \
 	aws iam delete-access-key --user-name <% .Name %>-ci-user --access-key-id $(shell aws iam list-access-keys --user-name <% .Name %>-ci-user --query "AccessKeyMetadata[0].AccessKeyId" | sed 's/"//g') && \
 	aws iam delete-user --user-name <% .Name %>-ci-user
 
+disable-rds-projection:
```
davidcheung, line 46: I'm not sure if adding this to `teardown-env` is a good idea, but I have had to go to the console to do this manually many times while trying, but if someone manually runs this then i think its okay

bmonkman: Yeah I wouldn't include this here, it's way too dangerous.

bmonkman: Oh this is just removing the protection. That should be fine, but I'm still a bit wary about adding too many commands in here that are only useful for us for testing, and won't ever be used in real world scenarios.

davidcheung: ill remove it, it really is mostly for testing convenience, for real users I guess they wont have an issue going to the web-console to do it manually

PR #48, opened 2020-06-26
ZERO138 update module name, add summary to makefile

cjhawkins:


zero-module.yml
```diff
@@ -1,4 +1,4 @@
-name: aws-eks-stack
```
davidcheung, line 1: 👍  didn't realize the other modules depends on `zero-aws-eks-stack`, was wondering why they weren't running haha

Makefile
```diff
@@ -2,3 +2,10 @@
 run:
 	cd $(PROJECT_DIR) && AUTO_APPROVE="-auto-approve" make
 
+summary:
+	@echo "zero-aws-eks-stack:"
+	@echo "- Repository URL: ${REPOSITORY}"
+	@echo "- To see your kubernetes clusters, run: 'kubectl config get-contexts'"
+	@echo "- To switch to a cluster, use the NAME from the previous command in 'kubectl config use-context NAME'"
```
bmonkman, line 9: Actually why don't we just show them the correct cluster name here?

cjhawkins: You want to see separate `kubectl config use-context`'s for each environment that is applied?

bmonkman: Yeah

PR #52, opened 2020-07-09
ZERO138 show commands to select each created cluster

cjhawkins:


Makefile
```diff
@@ -6,6 +6,7 @@ summary:
 	@echo "zero-aws-eks-stack:"
 	@echo "- Repository URL: ${REPOSITORY}"
 	@echo "- To see your kubernetes clusters, run: 'kubectl config get-contexts'"
-	@echo "- To switch to a cluster, use the NAME from the previous command in 'kubectl config use-context NAME'"
+	@echo "- To switch to a cluster, use the following commands:"
+	@echo $(shell echo ${ENVIRONMENT} | grep production > /dev/null && echo "- for production use: kubectl config use-context $(shell kubectl config get-contexts -o name | grep ${NAME}-production)")
```
bmonkman, line 10: A cleaner way to do this comparison might be something like:
`[[ ${ENVIRONMENT} =~ "production" ]] && echo "..."`

Makefile
```diff
@@ -1,3 +1,4 @@
+SHELL := /bin/bash
```
cjhawkins, line 1: I needed this to be able to use `[[` with Linux.

Makefile
```diff
@@ -6,6 +7,11 @@ summary:
 	@echo "zero-aws-eks-stack:"
 	@echo "- Repository URL: ${REPOSITORY}"
 	@echo "- To see your kubernetes clusters, run: 'kubectl config get-contexts'"
-	@echo "- To switch to a cluster, use the NAME from the previous command in 'kubectl config use-context NAME'"
+	@echo "- To switch to a cluster, use the following commands:"
+	@if [[ "${ENVIRONMENT}" =~ "production" ]]; then \
```
bmonkman, line 11: Can't you do it like this instead of using the long-form if?
`@echo $(shell [[ ${ENVIRONMENT} =~ "production" ]] && echo "...")`

cjhawkins: I'll give that a try.

cjhawkins: I got the short form working (shown below) but it will leave a blank space if the environment it is checking for was not applied.  Is that going to be OK?
`@echo $(shell [[ "${ENVIRONMENT}" =~ "production" ]] && echo "switch to production cluster: kubectl config use-context $(shell kubectl config get-contexts -o name | grep ${NAME}-production)")
`

bmonkman: Sure, that's fine for now. Thanks!

bmonkman: It's possible it would work with `@echo -n` but I wouldn't worry about it.

cjhawkins: I tried using `-n` but that also removed the newline in the case where there was text to show.  Also tried using `printf` with a `\n` but that didn't work either.   I'm going to merge it as is but I'll keep trying to find a better way to do this.

PR #55, opened 2020-07-31
Teardown improvement

davidcheung:
previously we had secret moved into `env` so it can be referenced as a resource with the unknown name-prefix timestamp

moving it back to secrets


templates/Makefile
```diff
@@ -40,6 +40,7 @@ teardown-remote-state:
 teardown-secrets:
```
bmonkman, line 40: What do you think about prompting the user if they are sure before running any of the teardown steps?

davidcheung: yeah perhaps a good idea since the terraform steps warn them before doing so 🤔

PR #60, opened 2020-08-12
Terraform013

davidcheung:
- upgrading terraform to 0.13
- adding mysql as a database engine option

### DIFF
--- a/templates/terraform/modules/database/main.tf
+++ b/templates/terraform/modules/database/main.tf
@@ -9,13 +9,19 @@ module "rds_security_group" {
 
   number_of_computed_ingress_with_source_security_group_id = 1
   computed_ingress_with_source_security_group_id = [
-    {
+    var.database == "postgres" ? {
       from_port   = 5432
       to_port     = 5432
       protocol    = "tcp"
       description = "PostgreSQL from EKS"
       source_security_group_id = "${var.allowed_security_group_id}"
-    },
+    }:{
+      from_port   = 3306
+      to_port     = 3306
+      protocol    = "tcp"
+      description = "MYSQL from EKS"
+      source_security_group_id = "${var.allowed_security_group_id}"
+    }
   ]
 
   egress_rules        = ["all-all"]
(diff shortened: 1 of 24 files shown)


templates/terraform/modules/database/main.tf
```diff
@@ -16,6 +16,13 @@ module "rds_security_group" {
       description = "PostgreSQL from EKS"
       source_security_group_id = "${var.allowed_security_group_id}"
     },
+    {
```
bmonkman, line 19: How about making this conditional so we aren't opening up a port they aren't using?

templates/terraform/modules/environment/variables.tf
```diff
@@ -63,3 +63,8 @@ variable "vpc_use_single_nat_gateway" {
   type        = bool
   default     = true
 }
+
+variable "database" {
+  default = "postgres"
+  description = "Which database driver to use, currently only supports postgres or mysql"
```
bmonkman, line 69: Should be "engine" or just "which database", not "driver"

templates/terraform/modules/database/main.tf
davidcheung: unless i group both of them in another layer of module, it seems like terraform doesn't like references to conditional modules, i get this error when have 2 conditional `rds_security_group`

```sh 
Error: Unsupported attribute

  on ../../modules/database/main.tf line 105, in module "rds_postgres":
 105:   vpc_security_group_ids = ["${module.rds_security_group_postgres.this_security_group_id}"]

This value does not have any attributes.


Error: Unsupported attribute

  on ../../modules/database/main.tf line 155, in module "rds_mysql":
 155:   vpc_security_group_ids = [module.rds_security_group_mysql.this_security_group_id]

This value does not have any attributes.
```

bmonkman: How about just having a single group but changing the rule depending on which db is chosen?

davidcheung: like this? https://github.com/commitdev/zero-aws-eks-stack/pull/60/commits/457fdc9663fb8c667cbe8e46e18434010407d254#diff-fc95e87f6973cfeeb3e12756fd64f5d5L10-L25

yeah this seems to work, not sure if it's a little harder to read though

bmonkman: If we add another type of DB the ternary operator wouldn't work and we'd have to switch to a lookup or something which might be cleaner, but for now it should be fine.

PR #61, opened 2020-08-13
Zero landing page [zero-222]

### LINKED ISSUE #59: Aws certificate manager domains; tags validation.
`apply-env` will fail on creating `aws_cloudfront_distribution` if you have multiple buckets. 

a tags k/v values can't have a  "," char value see 👇

```shell
Error: error creating CloudFront Distribution: MalformedXML: 1 validation error detected: Value '2020-08-11 17:56:32 +0000 UTC,2020-08-11 18:58:22 +0000 UTC' at 'distributionConfigWithTags.tags.items.1.member.value' failed to satisfy constraint: Member must satisfy regular expression pattern: ^([\p{L}\p{Z}\p{N}_.:/=+\-@]*)$
	status code: 400, request id: ea5d23e8-f775-42fa-9c2d-fad39ea461fa
  on ../../modules/s3_hosting/main.tf line 69, in resource "aws_cloudfront_distribution" "client_assets_distribution":
  69: resource "aws_cloudfront_distribution" "client_assets_distribution" {
```

change join ` "," -> "-"`
https://github.com/commitdev/zero-aws-eks-stack/blob/47abc825392bd6747a8a61169d6c5ef978a5b216/templates/terraform/modules/s3_hosting/main.tf#L131

related to https://github.com/commitdev/zero/issues/222

dtoki:
- create a new bucket using the root domain name.
- fix domain tag delimiter.

closes #59 
related to commitdev/zero#222

### DIFF
--- a/templates/terraform/modules/s3_hosting/main.tf
+++ b/templates/terraform/modules/s3_hosting/main.tf
@@ -128,7 +128,8 @@ resource "aws_cloudfront_distribution" "client_assets_distribution" {
 
   # Reference the cert validations only so it becomes a dependency
   tags = {
-    "certs-validated" = join(",", var.certificate_validations)
+    # join with realy used char - easier delimiter to indentify;
+    "certs-validated" = join("@", var.certificate_validations)
   }
 }
 
--- a/templates/terraform/modules/s3_hosting/main.tf
+++ b/templates/terraform/modules/s3_hosting/main.tf
@@ -128,8 +128,7 @@ resource "aws_cloudfront_distribution" "client_assets_distribution" {
 
   # Reference the cert validations only so it becomes a dependency
   tags = {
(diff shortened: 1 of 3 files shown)


templates/terraform/modules/s3_hosting/main.tf
```diff
@@ -128,7 +128,8 @@ resource "aws_cloudfront_distribution" "client_assets_distribution" {
 
   # Reference the cert validations only so it becomes a dependency
   tags = {
-    "certs-validated" = join(",", var.certificate_validations)
+    # join with realy used char - easier delimiter to indentify;
```
bmonkman, line 131: A couple typos in here. Also how about just remove this comment and use a dash. The comment and the @ make it look like this has some special significance, but really people should pay attention only to the above comment which indicates that the value doesn't matter.

PR #71, opened 2020-08-25
typo fix on logging elasticsearch subnet count

davidcheung:


davidcheung:
Yeah happens to me a lot too, unless we have a `terraform plan` with some default values just for sanity, would solve some issues but not the more complex dependencies

PR #74, opened 2020-08-31
69 support irsa

sshi100:
Please help review @bmonkman

### DIFF
--- a/templates/kubernetes/terraform/environments/dev/application_iam_policy.tf
+++ b/templates/kubernetes/terraform/environments/dev/application_iam_policy.tf
@@ -1,0 +1,24 @@
+
+# define policy documents for applications
+# sample policies
+data "aws_iam_policy_document" "resource_access_app1" {
+  statement {
+    effect = "Allow"
+    actions = [
+      "ec2:Describe*",
+    ]
+    resources = ["arn:aws:ec2:::dev-*"]
+  }
+  # can be more statements here
+}
+
+data "aws_iam_policy_document" "resource_access_app2" {
+  statement {
+    effect = "Allow"
+    actions = [
+      "ec2:CreateTags",
+    ]
+    resources = ["arn:aws:ec2:::dev-*"]
+  }
+  # can be more statements here
+}
--- a/templates/kubernetes/terraform/environments/dev/application_iam_policy.tf
+++ b/templates/kubernetes/terraform/environments/dev/application_iam_policy.tf
@@ -1,22 +1,11 @@
 
-# define policy documents for applications
+# define policy documents for backend services
 # sample policies
-data "aws_iam_policy_document" "resource_access_app1" {
+data "aws_iam_policy_document" "resource_access_backendservice" {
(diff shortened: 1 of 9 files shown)


templates/kubernetes/terraform/modules/kubernetes/irsa.tf
```diff
@@ -0,0 +1,48 @@
+# IRSA support: allow application specific policy via service-account and role
+
+# application_policy_list is passed from main.tf as below:
+  # application_policy_list = [
+  #   {
+  #     service_account = "app1"
+  #     namespace       = "piggycloud-me"
```
bmonkman, line 7: Maybe just "my-app" instead of "piggycloud"? And just one map in the list to make it shorter?

templates/kubernetes/terraform/modules/kubernetes/irsa.tf
```diff
@@ -0,0 +1,48 @@
...
+  role_name                     = "${var.project}-k8s-${var.environment}-${var.application_policy_list[count.index].service_account}"
+  provider_url                  = replace(data.aws_eks_cluster.cluster.identity.0.oidc.0.issuer, "https://", "")
+  role_policy_arns              = [aws_iam_policy.irsa[count.index].arn]
+  oidc_fully_qualified_subjects = ["system:serviceaccount:${var.application_policy_list[count.index].namespace}:${var.application_policy_list[count.index].service_account}"]
+}
+
+# Create policies
+resource "aws_iam_policy" "irsa" {
+  count       = length(var.application_policy_list)
+  name_prefix = var.project
```
bmonkman, line 33: Should probably be project + environment in case someone has two environments in a single aws account.

sshi100: done

templates/kubernetes/terraform/modules/kubernetes/irsa.tf
sshi100: done

templates/kubernetes/terraform/environments/dev/application_iam_policy.tf
```diff
@@ -0,0 +1,24 @@
+
+# define policy documents for applications
+# sample policies
+data "aws_iam_policy_document" "resource_access_app1" {
```
bmonkman, line 4: In all these files can you just make it a single example? The second one doesn't really provide any benefit.

templates/kubernetes/terraform/environments/stage/main.tf
```diff
@@ -32,4 +32,19 @@ module "kubernetes" {
 
   # Logging configuration
   logging_type = "<% index .Params `loggingType` %>"
+
+  # Application policy list
+  application_policy_list = [
+    {
+      service_account = "app1"
```
bmonkman, line 39: How about we change this service account name to "backend-service" or something and then also change the zero-deployable-backend project to use this service account?

Issue #75, opened 2020-08-31
Split modules out into separate repo


bmonkman:
As discussed in this proposal: https://docs.google.com/document/d/111E-ud8a6IHSAdZ346hUPOBTtIX2vJ-XkoiJcm4Li6s

To make it easier for people to upgrade and add more clarity on how users should structure their own changes to terraform we should:

- [x] Split all modules except for `environment` out to a separate repo (a single repo for all, for now)
- [x] Change `environment` to reference the new modules
- [x] Add documentation and examples to show users where they should add custom terraform
- [ ] Review outputs of modules to make sure we are exposing enough from each to make them useful 
- [ ] Look into setting up auto-deployment of our modules to the terraform registry

bmonkman:
https://registry.terraform.io/modules/commitdev/zero/aws/0.0.1

PR #76, opened 2020-09-01
#75 : Use new terraform registry source for all modules except enviro…

bmonkman:
…nment


bmonkman:
Will remove all the local modules in a separate PR.

PR #79, opened 2020-09-02
sendgrid integration

davidcheung:
Order of things that happens 
- while `zero apply` it will run `scripts/init`(in the project_dir) with parameters [`sendgridApiKey`] along with other env 
- running `scripts/init.sh` results in 2 files being create [`sendgrid.auto.tfvars.json` / `sendgrid-secrets.auto.tfvars.json`] in the environment's root (stage/prod) 
- `sendgrid-secrets.auto.tfvars.json` contains the sendgrid-api-key; and gets `.gitignored` so it will not be commited
- `sendgrid.auto.tfvars.json` will contain the sendgrid_domain_id(unique identifier from sendgrid) / CNAMEs we need for verification, and `send_enabled:true` (to override default sendgrid is off) 
- when apply runs `make apply-env` it will already have the variables in the root, then it will create route53, then call API to complete the verification 

assumes https://github.com/commitdev/terraform-aws-zero/pull/1 to be released as `0.0.2`


Makefile
```diff
@@ -1,8 +1,12 @@
 SHELL := /bin/bash
 
-run:
+run: sendgrid-init
 	cd $(PROJECT_DIR) && AUTO_APPROVE="-auto-approve" make
 
+sendgrid-init:
```
bmonkman, line 6: Should this be in the templated makefile instead?

davidcheung: That's where i wanted to put it as well 
but I need the `zero apply`'s env with the `sendgridApiKey`

bmonkman: Maybe we could push that into an aws secret during the secret init step, then have the script pull it out?

davidcheung: Yeah lets do that instead 👍

templates/scripts/pre-env.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
```
bmonkman, line 1: Maybe we should do something similar to what I did in post-apply, and source a script that is specific to sendgrid, then we can optionally include it. We may want to make this support optional via a parameter.

templates/scripts/pre-env.sh
```diff
@@ -0,0 +1,32 @@
...
+
+if [ "${ENVIRONMENT}" == "stage" ]; then
+  SENDGRID_DOMAIN="${DOMAIN_PREFIX}<% index .Params `stagingHostRoot` %>"
+elif [ "${ENVIRONMENT}" == "prod" ]; then
+  SENDGRID_DOMAIN="${DOMAIN_PREFIX}<% index .Params `productionHostRoot` %>"
+fi
+
+if [ ! -z "${SENDGRID_API_KEY}" ]
+then
+  MATCHING_DOMAIN_COUNT=$(curl -XGET "${SENDGRID_ENDPOINT}whitelabel/domains" -H "Authorization:Bearer ${SENDGRID_API_KEY}" | jq -c "[ .[] | select(.domain == \"${SENDGRID_DOMAIN}\")] | length");
```
bmonkman, line 21: All this stuff is quite vulnerable to them making changes to their API, but it seems like it's the only option.. :feelsbadman:

templates/terraform/environments/prod/main.tf
```diff
@@ -53,4 +53,8 @@ module "prod" {
   <% if ne (index .Params `loggingType`) "kibana" %># <% end %>logging_es_instance_count = "2" # Must be a mulitple of the az count
   <% if ne (index .Params `loggingType`) "kibana" %># <% end %>logging_volume_size_in_gb = "50" # Maximum value is limited by the instance type
   # See https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/aes-limits.html
+
+  sendgrid_enabled = var.sendgrid_enabled
+  sendgrid_cname = var.sendgrid_cname
```
bmonkman, line 58: This should be templated rather than a var?

davidcheung: I'm relying on the script to generate tfvars.json that will look something like 
```js
//terraform/environments/stage/sendgrid.auto.tfvars.json
{
  "sendgrid_enabled": true,
  "sendgrid_domain_id": 9104854,
  "sendgrid_cname": [
    [
      "em7596.email.get0.dev",
      "u18291290.wl041.sendgrid.net"
    ],
    [
      "s1._domainkey.email.get0.dev",
      "s1.domainkey.u18291290.wl041.sendgrid.net"
    ],
    [
      "s2._domainkey.email.get0.dev",
      "s2.domainkey.u18291290.wl041.sendgrid.net"
    ]
  ]
}

// terraform/environments/stage/sendgrid-secrets.auto.tfvars.json
{"sendgrid_apikey": "SG.blahblah-secret"}

```

bmonkman:
Could use something in the readme about how to create a sendgrid account and an api key and have everything in the right state to start this process.

templates/terraform/environments/stage/variables.tf
```diff
@@ -0,0 +1,17 @@
+variable "sendgrid_enabled" {
+  description = "If enabled, creates route53 entries for domain authentication"
+  type  = bool
+  default = false
+}
+
+variable "sendgrid_cname" {
```
bmonkman, line 7: should be "sendgrid_cnames"

templates/scripts/pre-env.sh
davidcheung: Yeah hopefully they won't make breaking changes on their `/v3` api, then we would have the option to manually upgrade it 🤞

templates/terraform/environments/prod/main.tf
bmonkman: Oh yeah, I must have missed that. Hm, not sure how I feel about this.. I'll think a bit more about it.

bmonkman: Yeah, actually this would make the tf totally unusable.

davidcheung: with the variables and default values in the stage/prod, they will just go to the default (enabled:false) and empty values

templates/terraform/modules/environment/main.tf
```diff
@@ -118,3 +118,12 @@ module "logging" {
...
   create_service_role   = var.logging_create_service_role
 }
+
+module "sendgrid" {
+  source = "commitdev/zero/aws//modules/sendgrid"
+  version = "0.0.2"
+  count  = var.sendgrid_enabled ? 1 : 0
+
+  zone_name = var.domain_name
+  sendgrid_api_key_secret_name = var.sendgrid_api_key_secret_name
```
davidcheung, line 128: almost all the logic has been moved to https://github.com/commitdev/terraform-aws-zero/pull/1/commits/bd194a0e680a10b59c15aa6e1236d050e2d2b82d with the external-datasource 🎉

templates/terraform/bootstrap/secrets/main.tf
```diff
@@ -47,3 +47,11 @@ module "rds_master_secret_prod" {
   random_length = 32
   tags = map("rds", "${local.project}-prod")
 }
+
+module "sendgrid_api_key" {
+  source  = "../../modules/secret"
+  name = "${local.project}-sendgrid-<% index .Params `randomSeed` %>"
```
bmonkman, line 53: This assumes the same api key will be shared for both staging / production, which I think is probably fine for now.

zero-module.yml
```diff
@@ -71,6 +71,9 @@ parameters:
     options:
       - "cloudwatch"
       - "kibana"
+  - field: sendgridApiKey
+    label: API key to setup email integration
+    info: Signup at https://signup.sendgrid.com or create an API key from https://app.sendgrid.com/settings/api_keys. Sendgrid is an email delivery service enabling transactional email sending and more.
```
bmonkman, line 76: How about a note here that they can leave this blank to not set up sendgrid?

bmonkman (approved):

PR #86, opened 2020-09-08
zero-234 Image Hosting Support

GrooveStomp:
https://app.zenhub.com/workspace/o/commitdev/zero/issues/234

- Moves backend app namespace configuration to main.tf from database-service.tf.
- Adds fileuploads question to cli interview.
- Adds new terraform for the backend app to backend-service.tf.


superkhau (approved):

templates/kubernetes/terraform/modules/kubernetes/backend-service.tf
```diff
@@ -0,0 +1,17 @@
+data "aws_secretsmanager_secret_version" "cf_keypair" {
+  secret_id = data.aws_secretsmanager_secret.cf_keypair.id
+  value = jsondecode(data.aws_secretsmanager_secret_version.cf_keypair.secret_string)
+}
+
+resource "kubernetes_secret" "cf_keypair" {
+  metadata {
+    name = "cf_keypair"
```
davidcheung, line 8: k8s resource names doesn't allow underscores, needs to be dashes instead

templates/kubernetes/terraform/modules/kubernetes/backend-service.tf
```diff
@@ -0,0 +1,17 @@
+data "aws_secretsmanager_secret_version" "cf_keypair" {
+  secret_id = data.aws_secretsmanager_secret.cf_keypair.id
```
davidcheung, line 2: I remember you telling me users will have to manually generate the keypair 
we should probably put a comment here indicating that its manual, 

Question: when its created it goes straight into secret-manager? or it is something we can help them put in secret manager?

GrooveStomp: I'm a little unclear on this myself; great question!
It's probably a better experience to have them provide it locally to Zero then we create the secret.  That also removes the jankiness of having to format the secret in a compatible way.

bmonkman: Yeah we definitely need instructions here. Maybe for now we just have a little script that imports the keys and writes aws secrets?

zero-module.yml
```diff
@@ -80,3 +80,9 @@ parameters:
   - field: randomSeed
     label: Random seed that will be shared between projects to come up with deterministic resource names
     execute: uuidgen | head -c 8
+  - field: fileUploads
+    label: Enable file uploads?
```
bmonkman, line 84: Maybe a bit more here? "Enable file uploads using S3 and Cloudfront signed URLs? (Will require manual creation of a Cloudfront keypair in AWS)"

templates/kubernetes/terraform/modules/kubernetes/backend-service.tf
```diff
@@ -0,0 +1,17 @@
+data "aws_secretsmanager_secret_version" "cf_keypair" {
```
bmonkman, line 1: Mind changing the name of the file to `backend_service.tf` with an underscore? You probably based this off `database-service.tf` which I've been meaning to rename because it's the only one that uses a dash. 🤦

bmonkman: Actually if you feel like renaming the database one too, that would be nice! 😁

GrooveStomp: Sure!

zero-module.yml
GrooveStomp: I'm all for more descriptive labels. :+1:

GrooveStomp:
I'm not sure where to put documentation for running the new script.
I don't see an _obvious_ spot in either the repo README or the template README for it to go.

templates/scripts/import-cf-keypair.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+
+# Inputs:
+# keypair ID
+# keypair private key
```
bmonkman: The private key you get after creating it in AWS is a file in PEM format, it's probably most convenient to just have them provide a file name as the second arg and then down below we `cat` it and pipe to `tr`?

bmonkman: oops, haha, you already did that. I was just looking at the commits in order. 👍

bmonkman:
I think the readme at the root of `templates/` is probably a good place.

templates/kubernetes/terraform/modules/kubernetes/backend_service.tf
```diff
@@ -0,0 +1,17 @@
+data "aws_secretsmanager_secret_version" "cf_keypair" {
+  secret_id = data.aws_secretsmanager_secret.cf_keypair.id
+  value = jsondecode(data.aws_secretsmanager_secret_version.cf_keypair.secret_string)
```
bmonkman, line 3: This is a data source, it looks up the secret value based on the id you provided, so you don't need to specify a value here. You also need a `aws_secretsmanager_secret` data source. You can see an example in modules/database/main.tf

templates/scripts/import-cf-keypair.sh
```diff
@@ -0,0 +1,47 @@
...
+fi
+
+ID=$1
+SECRET=$(tr -d '\r' < $2 | awk '{printf "%s\\n", $0}')
+
+aws secretsmanager \
+    create-secret \
+    --name <% .Name %>_cf_keypair \
+    --region <% index .Params `region` %> \
+    --secret-string "{\"keypair_id\":\"${ID}\",\"keypair_secret\":\"${SECRET}\"}"
```
bmonkman, line 47: Maybe we can call this `private_key` instead of `keypair_secret`?

GrooveStomp: Yeah, no prob. I'll update the node be repo to reflect that.

templates/kubernetes/terraform/modules/kubernetes/backend_service.tf
GrooveStomp: Ah, correctomundo.

templates/kubernetes/terraform/modules/kubernetes/backend_service.tf
```diff
@@ -0,0 +1,24 @@
...
+    name = "cf-keypair"
+  }
+
+  data = {
+    keypair_id = local.cf_keypair_json["keypair_id"]
+    secret_key = local.cf_keypair_json["private_key"]
+  }
+
+  type = "Opaque"
+}
```
GrooveStomp, line 24: I _think_ this should work properly this time?
I was setting an "imaginary" _value_ attribute on the **secret_version** resource previously for the decoded json.  I don't think that would have done anything meaningful.
I'm attempting to set it to a local var here - which appears to be kosher.

templates/scripts/import-cf-keypair.sh
```diff
@@ -0,0 +1,47 @@
...
+fi
+
+ID=$1
+SECRET=$(tr -d '\r' < $2 | awk '{printf "%s\\n", $0}')
+
+aws secretsmanager \
+    create-secret \
+    --name <% .Name %>_cf_keypair \
+    --region <% index .Params `region` %> \
+    --secret-string "{\"keypair_id\":\"${ID}\",\"private_key\":\"${SECRET}\"}"
```
GrooveStomp, line 47: Updated to `private_key` as per suggestion.

templates/scripts/import-cf-keypair.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+
+PROGNAME=$(basename "$0")
+
+function usage() {
+    echo "Usage: ${PROGNAME} [ -h | --help ] ID FILE"
+    echo
+    echo "Imports the specified CF Keypair data into AWS SecretsManager."
```
GrooveStomp, line 8: Added the contextual comment to the help output.

templates/kubernetes/terraform/modules/kubernetes/backend_service.tf
bmonkman: Yeah, looks good! If you don't have a good testing environment I can test this stuff out for you.

templates/scripts/import-cf-keypair.sh
```diff
@@ -0,0 +1,47 @@
...
+            break
+            ;;
+        *)
+            break
+            ;;
+    esac
+done
+
+if [ "$#" -ne 2 ]; then
+    echo "Incorrect number of arguments"
```
bmonkman, line 36: How about executing the `usage` function here?

templates/kubernetes/terraform/modules/kubernetes/backend_service.tf
```diff
@@ -0,0 +1,24 @@
+data "aws_secretsmanager_secret" "by-name" {
+  name = "<% .Name %>_cf_keypair"
```
bmonkman, line 2: This should be `var.project` instead of the templated variable. As much as possible in the terraform we try to keep the templating in the environment main.tf file, and then pass those values as arguments to the modules.

templates/kubernetes/terraform/modules/kubernetes/backend_service.tf
```diff
@@ -0,0 +1,24 @@
+data "aws_secretsmanager_secret" "by-name" {
```
bmonkman, line 1: The name of this should be something like `cf_keypair` instead of `by-name`

templates/kubernetes/terraform/modules/kubernetes/backend_service.tf
```diff
@@ -0,0 +1,24 @@
...
+  name = "<% .Name %>_cf_keypair"
+}
+
+data "aws_secretsmanager_secret_version" "cf_keypair" {
+  secret_id = data.aws_secretsmanager_secret.<% .Name %>_cf_keypair.id
```
bmonkman, line 6: You don't need the project name here, you are referring to the resource defined above. So it should be `data.aws_secretsmanager_secret.cf_keypair.id`

templates/kubernetes/terraform/modules/kubernetes/backend_service.tf
bmonkman: Yep, after making the changes recommended above both the script and terraform work properly.

templates/kubernetes/terraform/modules/kubernetes/backend_service.tf
```diff
@@ -0,0 +1,24 @@
...
+}
+
+locals {
+  cf_keypair_json = jsondecode(data.aws_secretsmanager_secret_version.cf_keypair.secret_string)
+}
+
+resource "kubernetes_secret" "cf_keypair" {
+  metadata {
+    name = "cf-keypair"
```
bmonkman, line 15: This needs to specify the namespace to create the secret in. In this case it should be `namespace = kubernetes_namespace.app_namespace.metadata[0].name`

templates/kubernetes/terraform/modules/kubernetes/backend_service.tf
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  data = {
+    keypair_id = local.cf_keypair_json["keypair_id"]
+    secret_key = local.cf_keypair_json["private_key"]
```
bmonkman, line 20: Should be `private_key` in the k8s secret as well

templates/scripts/import-cf-keypair.sh
GrooveStomp: Yeah...
I've done that previously and in using other tools I've seen more verbose output like this.
I've no strong leaning either way.
If you're in favour of the full help text, I'll do that.

templates/kubernetes/terraform/modules/kubernetes/backend_service.tf
GrooveStomp: Ah, cool! Thanks!

templates/kubernetes/terraform/modules/kubernetes/backend_service.tf
GrooveStomp: Are you sure?
I looked at the docs closely and it doesn't seem to be that way:

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/secretsmanager_secret
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/secretsmanager_secret_version

I am completely new to this, however!

GrooveStomp: Ah, I guess they're using a made-up-name that also explains the purpose of the example they're implementing.
Sorry about that.
Easy to misread...

bmonkman: Yeah. The syntax of `data` and `resource` in terraform always looks like `data "<type of thing you are referencing>" "<the name you are giving the thing so you can refer to it later>"`

GrooveStomp: Makes sense it would be consistent. :-)

templates/scripts/import-cf-keypair.sh
bmonkman: Sure, might as well. The first thing I did when I tried the script was run it with no arguments and the only feedback was that that was the wrong number of arguments. Would be nice to just have the full help text there even though the next thing most people will do is add `--help` (Okay, for most people that's probably the first thing they would do, but still.. ;) )

templates/kubernetes/terraform/modules/kubernetes/backend_service.tf
GrooveStomp: Thanks for working with me on this. I'm sure it must have been painful on your end! :-)

bmonkman: Hehe nope, it's no problem!

templates/kubernetes/terraform/modules/kubernetes/backend_service.tf
```diff
@@ -0,0 +1,24 @@
...
+  secret_id = data.aws_secretsmanager_secret.cf_keypair.id
+}
+
+locals {
+  cf_keypair_json = jsondecode(data.aws_secretsmanager_secret_version.cf_keypair.secret_string)
+}
+
+resource "kubernetes_secret" "cf_keypair" {
+  metadata {
+    namespace = kubernetes_namespace.app_namespace.metadata[0].name
```
bmonkman, line 15: Ah I'm sorry, here I meant that it should have `namespace` in addition to `name`, not replacing `name`. The secret still needs a name.

GrooveStomp: Ah, OK. At this point I'm going a little robotic in applying feedback; so that's my mistake. Thanks!

bmonkman (approved):
Looking good!

PR #87, opened 2020-09-09
Bump eks version, fix url in ami comment, remove unnecessary message …

bmonkman:
…during apply

### DIFF
--- a/templates/terraform/environments/prod/main.tf
+++ b/templates/terraform/environments/prod/main.tf
@@ -24,13 +24,13 @@ module "prod" {
   ecr_repositories = [] # Should be created by the staging environment
 
   # EKS configuration
-  eks_cluster_version      = "1.16"
+  eks_cluster_version      = "1.17"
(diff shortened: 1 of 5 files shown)


templates/terraform/environments/prod/main.tf
```diff
@@ -24,13 +24,13 @@ module "prod" {
   ecr_repositories = [] # Should be created by the staging environment
 
   # EKS configuration
-  eks_cluster_version      = "1.16"
+  eks_cluster_version      = "1.17"
```
davidcheung, line 27: https://github.com/commitdev/zero-aws-eks-stack/blob/main/templates/terraform/modules/environment/main.tf#L30

can you help bump this to 0.0.2 as well

davidcheung (approved):

templates/terraform/environments/prod/main.tf
bmonkman: Yes! Thanks! I knew there was something else.

PR #99, opened 2020-09-18
add new user - more commits with modules

sshi100:
Together with https://github.com/commitdev/terraform-aws-zero/pull/6

### DIFF
--- a/templates/Makefile
+++ b/templates/Makefile
@@ -23,9 +23,9 @@ apply-secrets:
 	rm ./terraform.tfstate )
 
 apply-shared-env:
-       cd terraform/environments/shared; \
-       terraform init && \
-       terraform apply $(AUTO_APPROVE)
+	cd terraform/environments/shared; \
+	terraform init && \
+	terraform apply $(AUTO_APPROVE)
 
 apply-env:
 	cd terraform/environments/$(ENVIRONMENT); \
--- a/templates/Makefile
+++ b/templates/Makefile
@@ -1,6 +1,7 @@
 SHELL = /usr/bin/env bash
 ENVIRONMENT ?= stage
 PROJECT = <% .Name %>
+ROLE ?= admin
 export AWS_DEFAULT_REGION = <% index .Params `region` %>
 export AWS_PAGER =
 KUBE_CONTEXT := $(PROJECT)-$(ENVIRONMENT)-$(AWS_DEFAULT_REGION)
@@ -40,7 +41,7 @@ apply-k8s-utils:
 	terraform apply $(AUTO_APPROVE)
 
 update-k8s-conf:
-	aws eks --region $(AWS_DEFAULT_REGION) update-kubeconfig --role "arn:aws:iam::<% index .Params `accountId` %>:role/$(PROJECT)-kubernetes-admin-$(ENVIRONMENT)" --name $(KUBE_CONTEXT) --alias $(KUBE_CONTEXT)
+	aws eks --region $(AWS_DEFAULT_REGION) update-kubeconfig --role "arn:aws:iam::<% index .Params `accountId` %>:role/$(PROJECT)-kubernetes-$(ROLE)-$(ENVIRONMENT)" --name $(KUBE_CONTEXT) --alias $(KUBE_CONTEXT)
 
 post-apply-setup:
 	cd scripts && ENVIRONMENT=$(ENVIRONMENT) PROJECT=$(PROJECT) sh post-apply.sh
--- a/templates/Makefile
+++ b/templates/Makefile
@@ -51,7 +51,7 @@ update-k8s-conf:
 post-apply-setup:
 	cd scripts && ENVIRONMENT=$(ENVIRONMENT) PROJECT=$(PROJECT) sh post-apply.sh
 
-teardown: teardown-k8s-utils teardown-env teardown-secrets teardown-remote-state
+teardown: teardown-k8s-utils teardown-env teardown-shared-env teardown-secrets teardown-remote-state
 
 teardown-remote-state:
 	@echo "Deleting remote state is not reversible, are you sure you want to delete the resources? [y/N]:" ; read ans ; [ $${ans:-N} == "y" ] || exit 1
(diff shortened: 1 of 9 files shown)


templates/terraform/environments/prod/user_access.tf
```diff
@@ -0,0 +1,29 @@
+# define policy documents for developer policy
+data "aws_iam_policy_document" "developer_access" {
+  # EKS
+  statement {
+    effect = "Allow"
+    actions = [
+      "eks:List*",
+      "eks:Describe*"
+    ]
+    resources = ["arn:aws:eks:::cluster/piggycloud-me-stage-*"]
```
bmonkman, line 10: Shouldn't be "piggycloud"

sshi100: done

templates/terraform/README.md
```diff
@@ -102,7 +102,15 @@
  make update-k8s-conf
  ```
 
+## User Access
 
+You may want to add people for AWS and Kubernetes access.
+ 1. Change user list with role and environment assignment in environments/shared/main.tf
+ 2. Change user access policy in environments/<env>/user_access.tf
+After applied, people can switch to assigned k8s context, in the root of the project, run:
```
bmonkman, line 110: We should note that they will need to apply in shared, and in each environment they changed.

templates/Makefile
```diff
@@ -40,7 +41,7 @@ apply-k8s-utils:
 	terraform apply $(AUTO_APPROVE)
```
bmonkman, line 42: We will need a new make target to apply the shared environment.

templates/terraform/environments/prod/user_access.tf
```diff
@@ -0,0 +1,111 @@
+# define AWS policy documents for developer
+data "aws_iam_policy_document" "developer_access" {
+  # IAM
+  statement {
```
bmonkman, line 4: This should be moved to the module. Added a note there.

templates/terraform/environments/prod/user_access.tf
```diff
@@ -0,0 +1,111 @@
+# define AWS policy documents for developer
+data "aws_iam_policy_document" "developer_access" {
```
bmonkman, line 2: At some point we will want to revisit these policies, add stuff like enforcing MFA, etc. but this should be fine for now.

templates/terraform/modules/environment/main.tf
```diff
@@ -8,6 +8,31 @@ data "aws_iam_user" "ci_user" {
...
+}
+
+locals {
+  role_name_list = var.roles.*.name
+  users = [
+    for u in data.terraform_remote_state.shared.outputs.user_role_mapping : {
+      name  = u.name
+      roles = [
+        for r in u.roles :
+        r.name if contains(local.role_name_list, r.name) && contains(r.environments, var.environment)
```
bmonkman, line 30: Nice, much cleaner now!

templates/terraform/modules/environment/main.tf
```diff
@@ -8,6 +8,31 @@ data "aws_iam_user" "ci_user" {
   user_name = "${var.project}-ci-user" # Should have been created in the bootstrap process
 }
 
+# get users from remote state of "shared"
+data "terraform_remote_state" "shared" {
```
bmonkman, line 12: Let's move this declaration out to the root and then pass the user role mapping in to the environment module? I don't like having the state reference hidden in a module (even if it is a local module..)

sshi100: move to the root and applied corresponding changes

templates/terraform/environments/prod/user_access.tf
sshi100: done

templates/Makefile
sshi100: done apply-shared-env prior to apply-env

templates/terraform/README.md
sshi100: done

templates/terraform/environments/prod/user_access.tf
```diff
@@ -0,0 +1,101 @@
...
+      "ecr:DescribeRepositories"
+    ]
+    resources = ["*"]
+  }
+
+  # S3
+  statement {
+    effect    = "Allow"
+    actions   = ["s3:ListBucket"]
+    resources = ["arn:aws:s3:::*<% index .Params `productionHostRoot` %>"]
```
bmonkman, line 29: Hmm if we're going to be using these values a lot more at the root how about putting them into locals in `main.tf`?

templates/terraform/environments/shared/main.tf
```diff
@@ -0,0 +1,60 @@
...
+
+# Instantiate the environment
+locals {
+  # Project configuration
+  project = "<% .Name %>"
+
+  # Users configuration
+  users = [
+    {
+      name  = "dev1"
```
bmonkman, line 25: Let's comment out these users, we don't actually want to create them.

templates/terraform/environments/stage/main.tf
```diff
@@ -35,15 +54,15 @@ module "stage" {
   eks_worker_asg_max_size  = 3
 
   # EKS-Optimized AMI for your region: https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html
-  # https://<% index .Params `region` %>.console.aws.amazon.com/systems-manager/parameters/%252Faws%252Fservice%252Feks%252Foptimized-ami%252F1.17%252Famazon-linux-2%252Frecommended%252Fimage_id/description?region=<% index .Params `region` %>
+  # https://${local.region}.console.aws.amazon.com/systems-manager/parameters/%252Faws%252Fservice%252Feks%252Foptimized-ami%252F1.17%252Famazon-linux-2%252Frecommended%252Fimage_id/description?region=${local.region}
```
bmonkman, line 57: This has to be templated in, it's in a comment.

templates/terraform/modules/environment/main.tf
```diff
@@ -28,7 +41,7 @@ data "aws_caller_identity" "current" {}
 # Provision the EKS cluster
 module "eks" {
   source  = "commitdev/zero/aws//modules/eks"
-  version = "0.0.2"
+  version = "0.1.2"
```
bmonkman, line 44: Could you please bump db and logging up to 0.1.2 as well?

bmonkman (approved):

PR #100, opened 2020-09-18
fix: syntax to support older bash versions

davidcheung:
the pipes confuses older bash versions to run commands after
pipe locally instead of passing to remote

with help from @sshi100 we found out in `bash 3.x` it gets confused and tries to run `wg` locally
and in `bash 5.x` it works fine

### DIFF
--- a/templates/scripts/add-vpn-user.sh
+++ b/templates/scripts/add-vpn-user.sh
@@ -70,7 +70,7 @@ Address = $next_ip/32
 # VPN server side
 PublicKey = $server_public_key
 AllowedIPs = 0.0.0.0/0
-Endpoint = vpn.$EXTERNAL_DNS:51820
+Endpoint = $EXTERNAL_DNS:51820
 
 EOF
 
--- a/templates/scripts/add-vpn-user.sh
+++ b/templates/scripts/add-vpn-user.sh
@@ -69,7 +69,7 @@ Address = $next_ip/32
 [Peer]
 # VPN server side
 PublicKey = $server_public_key
-AllowedIPs = 0.0.0.0/0
+AllowedIPs = 10.10.0.0/16
 Endpoint = $EXTERNAL_DNS:51820
 
 EOF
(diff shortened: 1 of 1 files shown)


bmonkman (approved):

templates/scripts/add-vpn-user.sh
```diff
@@ -6,24 +6,25 @@ CLUSTER=$(kubectl config current-context | cut -d"/" -f2)
 
 # get pod id for execution
 POD=$(kubectl -n vpn get pods | grep wireguard | cut -d' ' -f1)
+EXTERNAL_DNS=$(kubectl -nvpn get svc wireguard -o jsonpath='{.metadata.annotations.external-dns\.alpha\.kubernetes\.io/hostname}')
```
bmonkman, line 9: 👍

templates/scripts/add-vpn-user.sh
```diff
@@ -6,25 +6,26 @@ CLUSTER=$(kubectl config current-context | cut -d"/" -f2)
...
-client_public_key=$($EXEC echo -n $client_private_key | wg pubkey)
+server_public_key=$($EXEC "cat /etc/wireguard/privatekey | wg pubkey")
+client_private_key=$($EXEC "wg genkey")
+client_public_key=$($EXEC "echo -n $client_private_key | wg pubkey")
 
 # get next available IP
-existing_ips=$($EXEC cat /etc/wireguard/wg0.conf | grep AllowedIPs| cut -d" " -f3 | cut -d"/" -f1 | sort)
-last_ip=$(echo "$existing_ips" | tail -1)
+existing_ips=$($EXEC "cat /etc/wireguard/wg0.conf | grep AllowedIPs| cut -d\" \" -f3 | cut -d\"/\" -f1 | sort")
+last_ip=$(echo "$existing_ips" | tr -cd "[:alnum:]." | tail -1)
```
davidcheung, line 28: somehow when it doesnt work for me unless I sanitize the string `tr -cd "[:alnum:]."` 
I've tried to pipe it into a file and look at it but it looks fine

bmonkman (approved):

PR #104, opened 2020-09-21
Small fix to vpn output - carriage return was messing up the formatting

bmonkman:
After running the command my output looked like this:
```
"]  ["bill", "10.10.199.203/32", "asdf
```

Now it should look like this:
```
    ["bill", "10.10.199.203/32", "asdf"],
```


davidcheung:
on line 28 i added `tr -cd "[:alnum:]."` but really should be `tr -cd "[:alnum:].\n"`, or else it squashes all the lines together, can you help fix as well

davidcheung (approved):

PR #105, opened 2020-09-21
docs: sendgrid example and explaination

davidcheung:
this may not be super practical since most users will incorporate mail
into their application code, so the block is more explaining the state of
resources, indicating it is ready to send an email

### DIFF
--- a/templates/README.md
+++ b/templates/README.md
@@ -73,6 +73,28 @@ Please see [Link][zero-k8s-guide]
 - AWS IAM integration with Kubernetes RBAC
 ...
 
+#### Sending Email with Sendgrid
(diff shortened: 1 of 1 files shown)


templates/README.md
```diff
@@ -73,6 +73,28 @@ Please see [Link][zero-k8s-guide]
 - AWS IAM integration with Kubernetes RBAC
 ...
 
+#### Sending Email with Sendgrid
+Setup: If you initialized your infrastructure with a sendgridApiKey, you should have a verified domain with Sendgrid once your infrastructure is setup. We should have created for you 3 route53 entries, and [verified you domain with sendgrid][sendgrid-domain-verification]
```
bmonkman, line 77: Should be "We should have **created 3 route53 entries for you** and verified **your** domain with sendgrid"

bmonkman (approved):

PR #110, opened 2020-09-26
Lint status badge

davidcheung:
Now the init / validate steps are separate, 
but note that `init` will still catch some syntax errors
but validation will catch things like type validation (declared as list but passed in string)

image


README.md
```diff
@@ -1,4 +1,6 @@
 # Overview
+[![Validation Status](https://github.com/commitdev/zero-aws-eks-stack/workflows/Validate%20Terraform/badge.svg)](https://github.com/commitdev//zero-aws-eks-stack/actions)
```
bmonkman, line 2: extra slash here but the url still works.

bmonkman (approved):

PR #111, opened 2020-09-28
support DNS for VPN client

sshi100:


templates/scripts/add-vpn-user.sh
```diff
@@ -15,8 +17,7 @@ fi
 EXEC="kubectl -n vpn exec -it $POD -- /bin/bash -c"
 
 # get name
-echo -n "Enter your name: "
-read name
+echo -n "Enter your name: " && read name
```
bmonkman, line 20: Let's print the name of the cluster before this so they know which context they are using. Otherwise they don't really know where they are adding the user to, and I don't think we explicitly tell them to change their kube context.

templates/scripts/add-vpn-user.sh
```diff
@@ -31,6 +32,21 @@ while [[ "$existing_ips" =~ "$next_ip" ]]; do
...
+# get DNS server setting
+dns_server=$($EXEC "cat /etc/resolv.conf | grep nameserver | tail -1 | cut -d\" \" -f2 | tr -d \"\r\n\f\"")
+
+# get VPC CIDR for allowed IP subnet
+VPCNAME=${CLUSTER%-$REGION}-vpc
+vpc_cidr=$(aws ec2 describe-vpcs --filters Name=tag:Name,Values=${VPCNAME} | jq -r '.Vpcs[].CidrBlock')
+[[ -z "$vpc_cidr" ]] && vpc_cidr = "10.10.0.0/16"
+
+# get DB server
+K8S_DBSERVER=database.${NAMESPACE}.svc.cluster.local
```
bmonkman, line 44: Let's take this out. It's extra complexity just to show an example comment. Let's just leave the comment as an example like <rds mysql hostname>

templates/scripts/add-vpn-user.sh
```diff
@@ -50,7 +66,11 @@ echo
 echo "After this is done you should be able to open the wireguard client and activate the tunnel."
 echo "You can download the client at https://www.wireguard.com/install/"
 echo
-echo "When it is running you should be able to access internal resources, e.g. mysql -h 10.10.10.123"
+echo "When it is running you should be able to access internal resources, eg. database server:" 
+echo "  with Kubernetes cluster DNS, run:"
```
bmonkman, line 70: Let's not show both of these, it will just confuse people who don't have enough context. We can just have an example like <aws rds hostname>
If anything, you could mention that you can connect to things inside both the VPC and the kubernetes cluster.

templates/scripts/add-vpn-user.sh
```diff
@@ -65,11 +85,12 @@ cat <<-EOF > ${CONFIG_FILE}
 PrivateKey = $client_private_key
 ListenPort = 34567
 Address = $next_ip/32
+DNS = $dns_server
```
bmonkman, line 88: Isn't this just setting your dns to what was already set in your resolv.conf?

templates/scripts/add-vpn-user.sh
```diff
@@ -65,11 +85,12 @@ cat <<-EOF > ${CONFIG_FILE}
...
 
 [Peer]
 # VPN server side
 PublicKey = $server_public_key
-AllowedIPs = 10.10.0.0/16
+AllowedIPs = $vpc_cidr, $dns_server/32
```
bmonkman, line 93: Similar here, why would I want to route a request that should be on my local network through the vpn?

templates/scripts/add-vpn-user.sh
sshi100: Well, WireGuard client will handle it per OS. For MacOS by default, it doesn't change existing /etc/resolv.conf, but deal with mDNSResponder and the result can be found by running `scutil --dns`

bmonkman: Okay, didn't see that you were using the `$EXEC` var earlier to pull the dns from the pod.

sshi100: as discussed, that $dns_server is retrieved from vpn pod (not local)

templates/scripts/add-vpn-user.sh
sshi100: as confirmed, that $dns_server is retrieved from vpn pod (not from local)

templates/scripts/add-vpn-user.sh
sshi100: sure, good

templates/scripts/add-vpn-user.sh
sshi100: done

templates/scripts/add-vpn-user.sh
```diff
@@ -14,36 +14,38 @@ if [ -z "$POD" ]; then
   echo "Warning: No VPN service running yet"
   exit 1
 fi
-EXEC="kubectl -n vpn exec -it $POD -- /bin/bash -c"
+
+function k8s_exec() {
```
bmonkman, line 18: 👍

bmonkman (approved):

PR #116, opened 2020-09-29
Require MFA to be enabled for user accounts

### LINKED ISSUE #115: Require MFA for newly created IAM users

bmonkman:
(closes #115)

### DIFF
--- a/templates/terraform/environments/shared/main.tf
+++ b/templates/terraform/environments/shared/main.tf
@@ -54,6 +54,19 @@ resource "aws_iam_user" "access_user" {
   }
 }
 
+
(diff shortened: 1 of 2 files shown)


GrooveStomp (approved):
We don't assign users to any other IAM group?

bmonkman:
> We don't assign users to any other IAM group?

That happens in each environment (prod / stage), as those groups are per-environment, while this is global.
You can see the mapping, however, in `shared/main.tf`

sshi100 (approved):
I will add *Profile related for another ticket to enable user to access Console.

PR #120, opened 2020-09-30
118 console mngt for new user

sshi100:


templates/terraform/environments/shared/console_allowed.tf
```diff
@@ -0,0 +1,32 @@
...
+      "iam:GetLoginProfile",
+      "iam:CreateLoginProfile",
+      "iam:DeleteLoginProfile",
+      "iam:UpdateLoginProfile",
+    ]
+    resources = ["arn:aws:iam::${local.account_id}:user/$${aws:username}"]
+  }
+}
+
+resource "aws_iam_group" "console_allowed" {
```
bmonkman, line 24: This feels like too granular of a group, and something we should just be attaching to users by default but if you want to make it a group that's fine.

bmonkman (approved):

PR #124, opened 2020-09-30
Bug: K8s CF keypairs should be conditionally created

GrooveStomp:
https://app.zenhub.com/workspaces/commit-zero-5da8decc7046a60001c6db44/issues/commitdev/zero/295

- Create new var: cf_signing_enabled
- Conditionally lookup secret based on cf_signing_enabled
- Set count of k8s secret resources based on cf_sigining_enabled

### DIFF
--- a/templates/kubernetes/terraform/modules/kubernetes/fileupload/main.tf
+++ b/templates/kubernetes/terraform/modules/kubernetes/fileupload/main.tf
@@ -15,7 +15,7 @@ locals {
 resource "kubernetes_secret" "cf_keypair" {
   metadata {
     name = "cf-keypair"
-    namespace = kubernetes_namespace.app_namespace.metadata[0].name
+    namespace = var.namespace
   }
 
   data = {
--- a/templates/kubernetes/terraform/modules/kubernetes/fileupload/main.tf
+++ b/templates/kubernetes/terraform/modules/kubernetes/fileupload/main.tf
@@ -1,0 +1,27 @@
+# File uploads
+
+data "aws_secretsmanager_secret" "cf_keypair" {
+  name = "${var.project}_cf_keypair"
+}
+
+data "aws_secretsmanager_secret_version" "cf_keypair" {
+  secret_id = data.aws_secretsmanager_secret.cf_keypair.id
+}
+
+locals {
+  cf_keypair_json = jsondecode(data.aws_secretsmanager_secret_version.cf_keypair.secret_string)
+}
+
+resource "kubernetes_secret" "cf_keypair" {
+  metadata {
+    name = "cf-keypair"
+    namespace = kubernetes_namespace.app_namespace.metadata[0].name
+  }
+
+  data = {
+    keypair_id = local.cf_keypair_json["keypair_id"]
+    private_key = local.cf_keypair_json["private_key"]
+  }
+
+  type = "Opaque"
+}
--- a/templates/kubernetes/terraform/environments/prod/main.tf
+++ b/templates/kubernetes/terraform/environments/prod/main.tf
@@ -16,10 +16,11 @@ terraform {
 }
 
 locals {
-  project     = "<% .Name %>"
-  region      = "<% index .Params `region` %>"
-  account_id  = "<% index .Params `accountId` %>"
-  domain_name = "<% index .Params `productionHostRoot` %>"
+  project      = "<% .Name %>"
+  region       = "<% index .Params `region` %>"
+  account_id   = "<% index .Params `accountId` %>"
+  domain_name  = "<% index .Params `productionHostRoot` %>"
+  file_uploads = <% if eq (index .Params `fileUploads`) "yes" %>true<% else %>false<% end %>
 }
 
 provider "aws" {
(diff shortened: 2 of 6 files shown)


GrooveStomp:
How is this approach? I have yet to test it.

templates/kubernetes/terraform/environments/dev/main.tf
```diff
@@ -19,6 +19,8 @@ module "kubernetes" {
 
   project = "<% .Name %>"
 
+  cf_signing_enabled = "<% index .Params `fileUploads` %>"
```
davidcheung, line 22: i think the params stores it as a string so you may need some type casting -> bool

bmonkman:
Woo! The tests worked! 🎉

templates/kubernetes/terraform/environments/dev/main.tf
bmonkman: Yeah, this caused the tests to fail because "yes" and "no" are not valid bool values.

davidcheung:
another approach would be make a folder such as 
`kubernetes/terraform/modules/kubernetes/fileupload/main.tf` and have the current content of `backend-services.tf`
then in backend services 
have 
```
module "fileupload" {
  count = var.cf_signing_enabled ? 1 : 0
  source                     = "./fileupload"
  ...  some variables
}
```

templates/kubernetes/terraform/environments/dev/main.tf
GrooveStomp: > another approach would be make a folder such as
> `kubernetes/terraform/modules/kubernetes/fileupload/main.tf` and have the current content of `backend-services.tf`
> 
> then in `kubernetes/terraform/modules/kubernetes/main.tf` have
> 
> ```
> module "fileupload" {
>   count = var.cf_signing_enabled ? 1 : 0
>   source = "./fileupload"
>   ...  some variables
> }
> ```

Oooh, I like this.

GrooveStomp:
> another approach would be make a folder such as
> `kubernetes/terraform/modules/kubernetes/fileupload/main.tf` and have the current content of `backend-services.tf`
> 
> then in `kubernetes/terraform/modules/kubernetes/main.tf` have
> 
> ```
> module "fileupload" {
>   count = var.cf_signing_enabled ? 1 : 0
>   source = "./fileupload"
>   ...  some variables
> }
> ```

Ooh, I like this.

templates/kubernetes/terraform/environments/dev/main.tf
GrooveStomp: Ha ha, easy oversight.  I would have looked at the test _eventually_. ;-)

bmonkman:
I think that is overkill since we are only toggling a single resource. If there were a bunch of things to do a module might make sense, but not just for one resource.

davidcheung:
>I think that is overkill since we are only toggling a single resource. If there were a bunch of things to do a module might make sense, but not just for one resource.
theres some data source (the cf-secret itself)  and resources (k8s secret)

bmonkman:
Actually yeah, if the issue here is that the secret that we are referring to in the data source will conditionally not exist, then excluding it by putting it in a submodule is probably the only way. Otherwise I wouldn't recommend a submodule for this amount of code.

bmonkman:
I guess we could probably use `count` on the data source..

templates/kubernetes/terraform/modules/kubernetes/fileupload/main.tf
```diff
@@ -0,0 +1,27 @@
+# File uploads
+
+data "aws_secretsmanager_secret" "cf_keypair" {
+  name = "${var.project}_cf_keypair"
```
bmonkman, line 4: You need to define variables for this module.

GrooveStomp: I was hoping it would "pass through". :blush:

bmonkman: No, all modules are equal. Whether it's defined in a subdirectory, somewhere else on disk, or an external repo, they all have the same definition and requirements. You can think of them as functions, where input variables are the function signature and output variables are the return values. The only thing that is ever passed implicitly is providers.

GrooveStomp: Ah, that's what I was thinking of. Thanks for explaining it. :+1:

bmonkman (approved):

templates/kubernetes/terraform/modules/kubernetes/fileupload/main.tf
```diff
@@ -0,0 +1,27 @@
...
+}
+
+locals {
+  cf_keypair_json = jsondecode(data.aws_secretsmanager_secret_version.cf_keypair.secret_string)
+}
+
+resource "kubernetes_secret" "cf_keypair" {
+  metadata {
+    name = "cf-keypair"
+    namespace = kubernetes_namespace.app_namespace.metadata[0].name
```
bmonkman, line 18: Namespace will need to be passed in

davidcheung (approved):

PR #127, opened 2020-09-30
Some k8s tf cleanup.

bmonkman:
Deleted unused dev environment for now.
Moved re-used templating into locals in each env to match the other terraform.


GrooveStomp (approved):
So because `application_iam_policy.tf` and `main.tf` are in the same directory; they are part of the same module and can thus share local variables?

bmonkman:
They aren't part of a module at all, but yeah, whenever you have multiple `.tf` files in a directory, terraform ends up just merging them all together. Splitting them into mulitple files is just a way to organize things. There are popular conventions for things like modules like having `variables.tf`, `outputs.tf`, etc., but really you could just combine everything together into a single file if you wanted. The same behaviour applies when you are running `terraform plan/apply/etc` or for modules (notice that whenever you reference a module you point to a directory, not a file.)

PR #129, opened 2020-10-01
109 add user client setup

sshi100:
Changes:
- for Operator: `setup-user-password.sh` - setup user profile with temp password for new users (no profile & password setup yet), and output necessary info for Operator to pass to the users.
- for User: `setup-user-access.sh` - setup user local configurations for AWS and Kubernetes resource access
- minor fixes


templates/scripts/setup-user-access.sh
```diff
@@ -0,0 +1,98 @@
...
+  echo "  Usage:"
+  echo "    $0 <iam_username> <role> <environment>"
+  exit 1
+}
+
+function command_exist() {
+  command -v ${1} >& /dev/null
+}
+
+function confirm_aws_profile() {
```
bmonkman, line 31: If they are only called once and it's just a single line it doesn't make much sense to make them functions.. Just for cleanliness?

templates/terraform/environments/prod/main.tf
```diff
@@ -61,7 +61,7 @@ module "prod" {
   eks_worker_asg_max_size  = 4
 
   # EKS-Optimized AMI for your region: https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html
-  # https://${local.region}.console.aws.amazon.com/systems-manager/parameters/%252Faws%252Fservice%252Feks%252Foptimized-ami%252F1.17%252Famazon-linux-2%252Frecommended%252Fimage_id/description?region=${local.region}
+  # https://<% index .Params `region` %>.console.aws.amazon.com/systems-manager/parameters/%252Faws%252Fservice%252Feks%252Foptimized-ami%252F1.17%252Famazon-linux-2%252Frecommended%252Fimage_id/description?region=${local.region}
```
bmonkman, line 64: local.region is referred to again in this line.

templates/scripts/setup-user-password.sh
```diff
@@ -0,0 +1,95 @@
...
+      aws iam create-login-profile --user-name $user --password "$tpass" --password-reset-required >& /dev/null
+    fi
+    roles=$(aws iam list-groups-for-user --user-name $user | jq '.Groups[] | select(.Path == "/users/") | .GroupName' | tr -d "\"" | tr "\n" " ")
+    echo "  username:\"$tuser\", temporary password: \"$tpass\", roles: \"$roles\""
+}
+
+if [ -n "$INPUT_GROUP" ]; then 
+  nopass_users=$(aws iam get-group --group-name $INPUT_GROUP | jq ".Users[] | select(.PasswordLastUsed == null) | .UserName" | tr -d '"')
+
+  [[ -z ${nopass_users} ]] && warning_exit "No avaialbe users"
```
bmonkman, line 71: typo: "available"

templates/scripts/setup-user-password.sh
```diff
@@ -0,0 +1,95 @@
...
+[[ -z "$INPUT_GROUP" ]] && [[ -z "$INPUT_USER" ]] && INPUT_GROUP="console-allowed" # set default group
+
+[[ -n "$INPUT_GROUP" ]] && [[ -n "$INPUT_USER" ]] && warning_exit "Can not input both group and user togeher"
+
+# Collect users in this group without password
+function set_temp_password() {
+    user=$1
+    tpass=$(aws secretsmanager get-random-password --password-length 16 --require-each-included-type --output text)
+    aws iam get-login-profile --user-name $user >& /dev/null
+    if [ $? -eq 0 ]; then
```
bmonkman, line 59: If the user already has a password please just skip, we don't want to wipe out a legitimate password. Actually could you just do this check earlier and remove them from the list if they already have a login profile?

templates/scripts/setup-user-password.sh
```diff
@@ -0,0 +1,95 @@
...
+  done
+fi
+
+if [ -n "$INPUT_USER" ]; then
+  confirm "Will reset user \"$INPUT_USER\" password. Do you want to continue?" || warning_exit "No actions applied"
+  set_temp_password $INPUT_USER
+fi
+
+AWS_ACCOUNT_ID=<% index .Params `accountId` %>
+echo
```
bmonkman, line 91: A bit more instructions would be good here, like "You can send these credentials to each user via email or chat, they will be forced to choose a new password before being able to use their account."

templates/scripts/setup-user-password.sh
```diff
@@ -0,0 +1,95 @@
...
+    set_temp_password $tuser
+  done
+fi
+
+if [ -n "$INPUT_USER" ]; then
+  confirm "Will reset user \"$INPUT_USER\" password. Do you want to continue?" || warning_exit "No actions applied"
+  set_temp_password $INPUT_USER
+fi
+
+AWS_ACCOUNT_ID=<% index .Params `accountId` %>
```
bmonkman, line 90: Mind defining this at the top? It's kind of buried down here.

templates/scripts/setup-user-password.sh
```diff
@@ -0,0 +1,95 @@
...
+    roles=$(aws iam list-groups-for-user --user-name $user | jq '.Groups[] | select(.Path == "/users/") | .GroupName' | tr -d "\"" | tr "\n" " ")
+    echo "  username:\"$tuser\", temporary password: \"$tpass\", roles: \"$roles\""
+}
+
+if [ -n "$INPUT_GROUP" ]; then 
+  nopass_users=$(aws iam get-group --group-name $INPUT_GROUP | jq ".Users[] | select(.PasswordLastUsed == null) | .UserName" | tr -d '"')
+
+  [[ -z ${nopass_users} ]] && warning_exit "No avaialbe users"
+
+  echo "Detected the following users under group '$INPUT_GROUP':"
```
bmonkman, line 73: should be "in group"

templates/scripts/setup-user-password.sh
```diff
@@ -0,0 +1,95 @@
...
+  case $c in
+    h) usage ;;
+    g) INPUT_GROUP=$OPTARG ;;
+    u) INPUT_USER=$OPTARG ;;
+  esac
+done
+
+[[ -z "$INPUT_GROUP" ]] && [[ -z "$INPUT_USER" ]] && INPUT_GROUP="console-allowed" # set default group
+
+[[ -n "$INPUT_GROUP" ]] && [[ -n "$INPUT_USER" ]] && warning_exit "Can not input both group and user togeher"
```
bmonkman, line 52: Typo: "together"

templates/scripts/setup-user-password.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
```
bmonkman, line 1: How about calling this "create-temporary-passwords.sh" or something?

templates/scripts/setup-user-access.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
+
+#
+# This script is to help user to create local AWS and Kubernetes configurations
+# 
+# You should have the following information ready before running:
+# 1. your AWS user name
+# 2. your AWS access key
```
bmonkman, line 8: and secret

templates/scripts/setup-user-access.sh
```diff
@@ -0,0 +1,98 @@
...
+# 3. your user role
```
bmonkman, line 9: ("developer", "operator", etc.)

templates/scripts/setup-user-access.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
```
bmonkman, line 1: How about calling this something like "setup-local-aws-config"?

templates/scripts/setup-user-access.sh
```diff
@@ -0,0 +1,98 @@
...
+
+# templated variables
+AWS_ACCOUNT_ID=<% index .Params `accountId` %>
+PROJECT=<% .Name %>
+REGION=<% index .Params `region` %>
+
+# common functions
+function usage() {
+  echo "  Usage:"
+  echo "    $0 <iam_username> <role> <environment>"
```
bmonkman, line 23: Could you put some more info in here similar to the comment at the top?

bmonkman, line 23: "<environment (stage/prod)>"

templates/scripts/setup-user-access.sh
```diff
@@ -0,0 +1,98 @@
...
+if ! confirm_aws_identity ${MY_AWS_PROFILE}; then
+  warning_exit "Your profile not be able to setup, please check"
+fi
+echo "Confirmed your AWS profile '${MY_AWS_PROFILE}'"
+echo
+
+# Check & Configure K8S local context
+if ! command_exist kubectl
+then
+  warinign_exit "command 'kubectl' not found. You can download it at https://kubernetes.io/docs/tasks/tools/install-kubectl/."
```
bmonkman, line 81: typo in function name. Also please move this up to where we check for the aws command. Not a great experience to have to run this a bunch of times to find out the dependencies.
Maybe something like ```
if ! aws || !kubectl
  if !aws then warning
  if !kubectl then warning
  exit
fi
```

File: templates/scripts/setup-user-access.sh [hunk 15]
```diff
@@ -0,0 +1,98 @@
...
+
+NAMESPACE=${PROJECT}
+MY_K8S_CONTEXT=${PROJECT}-${ENVIRONMENT}-${REGION}
+if ! confirm_k8s_context ${MY_K8S_CONTEXT}; then
+  # setup or switch context
+  AWS_PROFILE=${MY_AWS_PROFILE} aws eks --region ${REGION} update-kubeconfig --role "arn:aws:iam::${AWS_ACCOUNT_ID}:role/${PROJECT}-kubernetes-${ROLE}-${ENVIRONMENT}" --name ${MY_K8S_CONTEXT} --alias ${MY_K8S_CONTEXT}
+  if ! confirm_k8s_context ${MY_K8S_CONTEXT}; then
+    warning_exit "Failed to setup your kubernetes context."
+  fi
+fi
```
bmonkman, line 92: How about switching their context here for them too?

templates/scripts/setup-user-access.sh
sshi100: did use it more than one time...

templates/terraform/environments/prod/main.tf
sshi100: okay, it's something back... okay

templates/scripts/setup-user-password.sh
sshi100: this makes sense... I thought to enforce this change if user wants it.

templates/scripts/setup-user-access.sh
sshi100: to "setup-local-config.sh"

templates/scripts/setup-user-access.sh
sshi100: This is a check after the context setup/switch done... so switch to something else?

templates/scripts/setup-user-access.sh
sshi100: let's keep it

templates/scripts/setup-user-password.sh
sshi100: Confirmed: the user list targets people whose .PasswordLastUsed == null, which means NoPasswordYet or Password expired, based on document explanation:
```aws iam list-users returns a list of users, each with a PasswordLastUsed value. If the value is missing, then the user either has no password or the password has not been used since IAM began tracking password age on October 20, 2014.```

So we don't wipe out a legitimate password.

File: templates/scripts/setup-user-access.sh (hunk 13 above)
Comment (sshi100) [reply templates/scripts/setup-user-access.sh:23]: sure

File: templates/scripts/setup-user-access.sh (hunk 11 above)
Comment (sshi100) [reply templates/scripts/setup-user-access.sh:9]: done

File: templates/scripts/setup-user-access.sh (hunk 10 above)
Comment (sshi100) [reply templates/scripts/setup-user-access.sh:8]: done

File: templates/scripts/setup-user-password.sh (hunk 9 above)
Comment (sshi100) [reply templates/scripts/setup-user-password.sh:1]: yes, use this "create-temporary-passwords.sh"

File: templates/scripts/setup-user-password.sh (hunk 6 above)
Comment (sshi100) [reply templates/scripts/setup-user-password.sh:90]: done

File: templates/scripts/setup-user-password.sh (hunk 5 above)
Comment (sshi100) [reply templates/scripts/setup-user-password.sh:91]: sure

File: templates/scripts/setup-user-password.sh (hunk 4 above)
Comment (sshi100) [reply templates/scripts/setup-user-password.sh:59]: after more studies and testings, found there are others cases around profile (yes/no) vs password(yes/no). So I think we can just remove it from the list if already have a login-profile.

File: templates/scripts/setup-user-access.sh (hunk 15 above)
Comment (sshi100) [reply templates/scripts/setup-user-access.sh]: per Slack chatting, it means we can do `kubectl set current-context`.

Comment (sshi100) [reply templates/scripts/setup-user-access.sh]: `aws eks update-kubeconfig` does the work

File: templates/scripts/create-temporary-passwords.sh [hunk 16]
```diff
@@ -0,0 +1,106 @@
...
+  case $c in
+    h) usage ;;
+    g) INPUT_GROUP=$OPTARG ;;
+    u) INPUT_USER=$OPTARG ;;
+  esac
+done
+
+[[ -z "$INPUT_GROUP" ]] && [[ -z "${INPUT_USER}" ]] && INPUT_GROUP="console-allowed" # set default group
+
+[[ -n "$INPUT_GROUP" ]] && [[ -n "${INPUT_USER}" ]] && warning_exit "Can not input both group and user togeher"
```
bmonkman, line 55: typo "togeher"

templates/scripts/setup-local-config.sh
```diff
@@ -0,0 +1,95 @@
...
+  echo "Setup new AWS profile: ${MY_AWS_PROFILE}"
+  read -p "  AWS Access Key: " aws_access_key
+  read -s -p "  AWS Secret Key:" aws_secret_key
+  echo
+  aws configure set --profile ${MY_AWS_PROFILE} aws_access_key_id $aws_access_key
+  aws configure set --profile ${MY_AWS_PROFILE} aws_secret_access_key $aws_secret_key
+  aws configure set --profile ${MY_AWS_PROFILE} region ${REGION}
+fi
+if ! confirm_aws_identity ${MY_AWS_PROFILE}; then
+  warning_exit "Your profile not be able to setup, please check"
```
bmonkman, line 77: How about something like "Unable to set up your AWS configuration profile. Please contact an administrator."

bmonkman (approved):

PR #130, opened 2020-10-02
improvement: specifying non-self trusted signers

davidcheung:


GrooveStomp (changes requested):
Looks good, but needs an updated version for `s3_hosting`.

GrooveStomp (approved):

PR #133, opened 2020-10-05
prepend protocol for S3 CORS origin

davidcheung:


templates/terraform/environments/prod/main.tf
```diff
@@ -17,10 +17,11 @@ terraform {
...
 locals {
-  project     = "<% .Name %>"
-  region      = "<% index .Params `region` %>"
-  account_id  = "<% index .Params `accountId` %>"
-  domain_name = "<% index .Params `productionHostRoot` %>"
+  project             = "<% .Name %>"
+  region              = "<% index .Params `region` %>"
+  account_id          = "<% index .Params `accountId` %>"
+  domain_name         = "<% index .Params `productionHostRoot` %>"
+  app_protocol_prefix = "https://"
```
bmonkman, line 24: I'm not sure this is useful exposing as a local here since it implies that you can change it. If they change it, it will just break. I would say hard code it below, and if they really need to add another domain should be able to refer traffic, they can just add another string into the `cors_origins` list with the correct scheme.

bmonkman (changes requested):

bmonkman (approved):

PR #134, opened 2020-10-06
initial cloudtrail commit

sshi100:


templates/terraform/environments/prod/main.tf
```diff
@@ -122,4 +122,8 @@ module "prod" {
   ]
 
   user_role_mapping = data.terraform_remote_state.shared.outputs.user_role_mapping
+
+  # Cloud trail configuration
+  cloudtrail_enabled = true
+  cloudtrail_include_global_service_events = false
```
bmonkman, line 128: How about a comment about what this means

templates/terraform/modules/environment/main.tf
```diff
@@ -162,3 +162,15 @@ module "user_access" {
   roles = var.roles
   users = local.users
 }
+
+module "cloudtrail" {
+  source = "commitdev/zero/aws//modules//cloudtrail"
```
bmonkman, line 167: The double slash should only be at the root of the project, like this: `commitdev/zero/aws//modules/cloudtrail`
Mind changing `user_access` too?

templates/terraform/environments/prod/main.tf
```diff
@@ -122,4 +122,8 @@ module "prod" {
   ]
 
   user_role_mapping = data.terraform_remote_state.shared.outputs.user_role_mapping
+
+  # Cloud trail configuration
```
bmonkman, line 126: Maybe a comment here about what cloudtrail does, why you might want it enabled, and the cost?

templates/terraform/modules/environment/main.tf
```diff
@@ -162,3 +162,15 @@ module "user_access" {
...
+  version = "0.1.4"
```
bmonkman, line 168: Reminder that this will have to be updated when the module is published.

bmonkman (approved):

PR #141, opened 2020-10-10
Add prometheus and grafana support, and new metricsType configuration

### LINKED ISSUE #136: Add Prometheus / Grafana to infrastructure repo
@bmonkman commented on [Wed Aug 19 2020](https://github.com/commitdev/zero/issues/250)

bmonkman:
Configured with prometheus-operator
Sets up an ingress, but we don't have proper internal domain support so it doesn't do much for now.

(Closes #136)﻿

### DIFF
--- a/templates/kubernetes/terraform/environments/prod/main.tf
+++ b/templates/kubernetes/terraform/environments/prod/main.tf
@@ -49,8 +49,10 @@ module "kubernetes" {
   # Registration email for LetsEncrypt
   cert_manager_acme_registration_email = "devops@${local.domain_name}"
 
-  # Logging configuration
+  # Logging and Metrics configuration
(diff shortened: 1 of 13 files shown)


templates/kubernetes/terraform/modules/kubernetes/metrics/prometheus/files/prometheus_operator_helm_config.yml
```diff
@@ -0,0 +1,2043 @@
+# Default values for prometheus-operator.
```
bmonkman, line 1: The values in here are 99% the defaults for the prometheus-operator helm chart. I'm a little torn on whether to leave this file with all the options in it, or to exclude everything that is a default value..

PR #142, opened 2020-10-13
140 use new cloudtrail module

sshi100:
Changes:
- create cloudtrail at `shared` instead of that creation per environment
- refine user access policy with bucket name from `shared` data source


sshi100:
close this pull request and redo

PR #145, opened 2020-10-13
added a fix of output name for cloudtrail

sshi100:


sshi100:
this output is used by user_access.tf

PR #147, opened 2020-10-18
285 make ci user with module user_access

sshi100:
Changes:
- move ci-user creation and access key handling out of `bootstrap` to `shared` and assign to new role `deployer`
- added new parameter `create_access_keys` etc. for roles
- set `deploye`r policy same as `operator`
- not allow ci-user to access console

### DIFF
--- a/templates/terraform/environments/prod/user_access.tf
+++ b/templates/terraform/environments/prod/user_access.tf
@@ -185,7 +185,8 @@ locals {
       api_groups = ["*"]
       resources = ["deployments", "configmaps", "pods", "pods/exec", "pods/log", "pods/status", "pods/portforward",
         "jobs", "cronjobs", "secrets", "services", "daemonsets", "endpoints", "namespaces", "events", "ingresses",
-        "horizontalpodautoscalers", "horizontalpodautoscalers/status"
+        "horizontalpodautoscalers", "horizontalpodautoscalers/status",
+        "poddisruptionbudgets"
       ]
     }
   ]
@@ -197,7 +198,8 @@ locals {
       api_groups = ["*"]
       resources = ["deployments", "configmaps", "pods", "pods/log", "pods/status",
         "jobs", "cronjobs", "services", "daemonsets", "endpoints", "namespaces", "events", "ingresses",
-        "horizontalpodautoscalers", "horizontalpodautoscalers/status"
+        "horizontalpodautoscalers", "horizontalpodautoscalers/status",
+        "poddisruptionbudgets"
       ]
     },
     {
--- a/templates/terraform/environments/prod/user_access.tf
+++ b/templates/terraform/environments/prod/user_access.tf
@@ -181,7 +181,7 @@ locals {
   # define Kubernetes policy for operator
   k8s_operator_access = [
     {
-      verbs      = ["exec", "create", "list", "get", "delete", "patch", "update"]
+      verbs      = ["exec", "create", "list", "get", "delete", "patch", "update", "watch"]
       api_groups = ["*"]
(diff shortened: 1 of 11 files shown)


templates/terraform/environments/prod/user_access.tf
```diff
@@ -110,4 +184,13 @@ locals {
       resources  = ["deployments", "configmaps", "pods", "secrets", "services", "endpoints"]
     }
   ]
+
+  # define Kubernetes policy for deployer: To Be Refined later
+  k8s_deployer_access = [
```
bmonkman, line 189: This isn't enough access to be able to deploy.

templates/terraform/environments/shared/main.tf
```diff
@@ -24,27 +25,40 @@ provider "aws" {
 locals {
   # Users configuration
   users = [
-    #    {
+        {
+          name  = "${local.project}-ci-user"
+          roles = [
+            { name = "deployer", environments = ["stage", "prod"] }
+          ]
+          global_roles       = ["console-allowed"]
```
bmonkman, line 33: Let's remove console access as well.

templates/terraform/modules/environment/main.tf
```diff
@@ -2,25 +2,35 @@
...
       roles = [
         for r in u.roles :
-        r.name if contains(local.role_name_list, r.name) && contains(r.environments, var.environment)
+        r.name if contains(var.roles.*.name, r.name) && contains(r.environments, var.environment)
       ]
     }
   ]
+
+  ci_users = [
+    for u in local.users: u.name if contains(u.roles, "deployer")
```
bmonkman, line 17: Let's not do it based on if there is "deployer" in the name. That's not at all obvious to people who are using this. Either let's target this specific user, or let's provide some sort of data in the user list that lets us know this is a special type  of user, if we actually need to treat them differently.

templates/terraform/modules/environment/main.tf
```diff
@@ -162,3 +172,14 @@ module "user_access" {
   roles = var.roles
   users = local.users
 }
+
+
+output "s3_hosting" {
+  value = [
```
bmonkman, line 178: Can you add a description please?

templates/terraform/environments/prod/user_access.tf
sshi100: make all * for the time being.

sshi100: set same as operator

templates/terraform/environments/prod/user_access.tf
```diff
@@ -115,4 +189,16 @@ locals {
       ]
     }
   ]
+
+  # define Kubernetes policy for deployer
+  k8s_deployer_access = [
+    {
+      verbs      = ["exec", "create", "list", "get", "delete", "patch", "update"]
+      api_groups = ["*"]
+      resources = ["deployments", "configmaps", "pods", "pods/exec", "pods/log", "pods/status", "pods/portforward",
```
bmonkman, line 198: Let's tweak this a bit to remove port forward and exec, and maybe make secrets write-only? 
Please take a look at the default deployment manifests to make sure this policy covers it. Otherwise even our demo apps won't be able to deploy. https://github.com/commitdev/zero-deployable-backend/tree/main/templates/kubernetes
(For example, one thing I think we create is a PodDisruptionBudget

sshi100: added "watch", set secrets write-only, removed portforward/exec etc.

bmonkman (approved):

PR #152, opened 2020-10-27
create default application DB user with TF

sshi100:
Changes:
- used mysql/postgres provider but gave up because that supposes TF can access to RDS server
- use k8 job with SQL script to create/update user account, and store to k8s secret for application to load at run time
- use password version with `random_password` to control if do recreation

PR #153, opened 2020-10-29
port from zero-deployable-backend-sevice and make enhancement

sshi100:
Instead of https://github.com/commitdev/zero-aws-eks-stack/pull/152, we decided to port original backend-service into infrastructure considering of security etc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,7 @@
 SHELL := /bin/bash
 
-run: create-db-user
+run:
 	cd $(PROJECT_DIR) && AUTO_APPROVE="-auto-approve" make
-
-create-db-user:
 	kubectl -n ${PROJECT_NAME} get secrets ${PROJECT_NAME} > /dev/null 2>&1 || ( \
 	export REGION=${region}; \
 	export SEED=${randomSeed}; \
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,16 @@
 SHELL := /bin/bash
 
-run:
+run: create-db-user
 	cd $(PROJECT_DIR) && AUTO_APPROVE="-auto-approve" make
+
+create-db-user:
+	kubectl -n ${PROJECT_NAME} get secrets ${PROJECT_NAME} > /dev/null 2>&1 || ( \
+	export REGION=${region}; \
+	export SEED=${randomSeed}; \
+	export PROJECT_NAME=${PROJECT_NAME}; \
+	export ENVIRONMENT=${ENVIRONMENT}; \
+	export DATABASE=${database}; \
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -1,8 +1,17 @@
 SHELL := /bin/bash
 
-run:
+run: create-db-user
```
bmonkman, line 3: This will try to create the user before creating the infrastructure. Could you make sure to test this while setting up a new project?

templates/README.md
```diff
@@ -90,6 +90,16 @@ $ curl --request POST \
...
 For Application use, see [Sendgrid resources][sendgrid-send-mail] on how to setup templates to send dynamic transactional emails. To setup emailing from your application deployment, you should create a kubernetes secret with your Sendgrid API Key(already stored in [AWS secret-manager](./terraform/bootstrap/secrets/main.tf)) in your application's namespace. Then mount the secret as an environment variable in your deployment.
 
+#### Application Database user creation
+Using environment variables injected from Zero, it will fetch the RDS master password from AWS secret manager
+- creates a namespace
+- creates a job with a SQL query file mounted generating an application user
+- creating a secret in the application namespace in your EKS cluster
+- removing the RDS master password for security reasons
+
+_Note: the user creation only happens once during `zero apply`, for details see the `make create-db-user` command. If you want to renew the user/password, you can run the script `sh dp-ops/create-db-user.sh` manually.
```
bmonkman, line 100:
```suggestion
#### Application database user creation
A database user will automatically be created for a backend application with a random password, and the credentials will be stored in a kubernetes secret so they are available to the application.

_Note: the user creation only happens once during `zero apply`. If you want to run this process again to create a new password, you can run the script `sh dp-ops/create-db-user.sh` manually, though be aware that this will cause the user and its privileges to be removed and recreated, and may disrupt your application until it is restarted.
```

sshi100:
All confirmed. In addition:
- fixed Makefile with describe-secret instead of list-secret
- fixed remaining interpolation-errors in TF

Please review @bmonkman

Makefile
```diff
@@ -1,9 +1,7 @@
 SHELL := /bin/bash
 
-run: create-db-user
+run:
 	cd $(PROJECT_DIR) && AUTO_APPROVE="-auto-approve" make
-
-create-db-user:
 	kubectl -n ${PROJECT_NAME} get secrets ${PROJECT_NAME} > /dev/null 2>&1 || ( \
```
bmonkman, line 5: Is the kubectl context set properly during make so that this actually works?

bmonkman: Should be, if it was working when run from another project.

bmonkman (approved):

PR #154, opened 2020-10-31
Initial commit of auth proxy and user management

bmonkman:
Using Ory Kratos and Oathkeeper.
Added Zero module parameter to enable or disable.
Still need to add proper DB and mail support.
Not including oathkeeper rules for now, as since we are using the operator we can define them in the backend project.


db-ops/job-create-db-mysql.yml.tpl
```diff
@@ -2,51 +2,45 @@ apiVersion: v1
...
 metadata:
   name: db-create-users
   namespace: db-ops
 type: Opaque
-stringData: 
+stringData:
+  RDS_MASTER_PASSWORD: $MASTER_RDS_PASSWORD
   create-user.sql: |
-    DROP USER '$DB_APP_USERNAME';
+    CREATE DATABASE IF NOT EXISTS $DB_NAME;
```
bmonkman, line 16: Creation would fail without these

db-ops/job-create-db-postgres.yml.tpl
```diff
@@ -2,52 +2,46 @@ apiVersion: v1
...
   name: db-create-users
   namespace: db-ops
 type: Opaque
-stringData: 
+stringData:
+  RDS_MASTER_PASSWORD: $MASTER_RDS_PASSWORD
   create-user.sql: |
-    REVOKE ALL PRIVILEGES ON DATABASE $DB_NAME FROM $DB_APP_USERNAME;
-    DROP USER $DB_APP_USERNAME;
+    SELECT 'CREATE DATABASE $DB_NAME' WHERE NOT EXISTS (SELECT FROM pg_database WHERE datname = '$DB_NAME');\gexec
```
bmonkman, line 16: Creation would fail without these

db-ops/job-create-db-postgres.yml.tpl
```diff
@@ -2,52 +2,46 @@ apiVersion: v1
...
         image: $DOCKER_IMAGE_TAG
-        command: 
+        command:
         - sh
-        args: 
-        - '-c' 
-        - psql -U$MASTER_RDS_USERNAME -h $DB_ENDPOINT $DB_NAME -a -f/db-ops/create-user.sql > /dev/null
+        args:
+        - '-c'
+        - psql -U$MASTER_RDS_USERNAME -h $DB_ENDPOINT postgres -v ON_ERROR_STOP=1 -a -f/db-ops/create-user.sql > /dev/null
```
bmonkman, line 44: This script would succeed even if there were errors without this

db-ops/job-create-db-mysql.yml.tpl
davidcheung: the terraform RDS module we use creates a database so it should be fine for most cases

templates/kubernetes/terraform/modules/kubernetes/files/kratos-values.yml
```diff
@@ -0,0 +1,160 @@
...
+            #   - id: github
+            #     provider: github
+            #     client_id: # TODO: load secret
+            #     client_secret: # TODO : load secret
+            #     # TODO: Change to static or deploy somewhere within infra
+            #     mapper_url: http://zombism.ca/github.data-mapper.jsonnet
+            #     scope:
+            #       - user:email
+
+      default_browser_return_url: /dashboard
```
davidcheung, line 99: we may have have to have the entire frontend domain in all of these paths or else it redirects to the backend instead of back to the UI

bmonkman: Yeah, need to tweak those for sure.

db-ops/job-create-db-mysql.yml.tpl
bmonkman: Yeah though if there is a dash in the name I'm removing it which I think means it might not exist.

davidcheung: yeah we have a line in the database module to do that too https://github.com/commitdev/terraform-aws-zero/blob/main/modules/database/main.tf#L61

sshi100 (approved):
good on refined "create DB user" part!

templates/kubernetes/terraform/modules/kubernetes/user_auth.tf
```diff
@@ -54,6 +54,11 @@ resource "helm_release" "oathkeeper" {
     value = var.backend_service_domain
   }
 
+  set {
+    name  = "ingress.proxy.annotations.nginx\\.ingress\\.kubernetes\\.io/cors-allow-origin"
+    value = "https://${var.backend_service_domain}"
```
davidcheung, line 59: should be frontend

davidcheung (approved):

PR #159, opened 2020-11-05
added k8s_groups for vpn-users

sshi100:
Changes:
- add `k8s_groups` parameter into user role, pass to module `eks`
- add `vpn-users` into `k8s_groups`
- create Role and RoleBinding to handle group `vpn-users`

### DIFF
--- a/templates/terraform/modules/environment/main.tf
+++ b/templates/terraform/modules/environment/main.tf
@@ -3,7 +3,7 @@
 locals {
   kubernetes_cluster_name = "${var.project}-${var.environment}-${var.region}"
 
-  users = [
+  user_access_users = [
     for u in var.user_role_mapping : {
       name = u.name
       roles = [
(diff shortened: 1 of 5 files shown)


templates/terraform/modules/environment/main.tf
```diff
@@ -12,12 +12,22 @@ locals {
...
+  ]
 
-  kubernetes_iam_role_mapping = [
+  eks_kubernetes_iam_role_mapping = [
     for r in module.user_access.eks_iam_role_mapping : {
       iam_role_arn  = r.arn
       k8s_role_name = r.name
-      k8s_groups    = [ r.name ]
+      k8s_groups    = flatten(concat([r.name], [
+        for o in var.roles : o.k8s_groups if r.name == "${var.project}-kubernetes-${o.name}-${var.environment}"
```
bmonkman, line 29: What's with this condition? Shouldn't we just be always merging `k8s_groups` with `r.name` ?

bmonkman (approved):

templates/terraform/modules/environment/main.tf
sshi100: combining 'module.user_access.eks_iam_role_mapping' and 'var.roles' to set 'k8s_groups'

PR #163, opened 2020-11-25
Maintenance and fixes

bmonkman:
Bumped versions of various helm charts, with related config fixes.
Moved db ops scripts into templates dir, call them from the templated makefile rather than the one at the root so we have more control.
Misc other fixes to support completing a full run of `zero apply`


templates/scripts/post-apply.sh
```diff
@@ -1,4 +1,20 @@
 #!/bin/sh
-set -e
+
+KUBE_CONTEXT=${PROJECT}-${ENVIRONMENT}-${AWS_DEFAULT_REGION}
```
davidcheung, line 3: I notice we're no longer using the region from `parameters` but using `AWS_DEFAULT_REGION` instead
would this cause issues where the user's AWS profile has a different default region than the one picked?

bmonkman: The makefile is setting `AWS_DEFAULT_REGION` to the content of the `region` parameter.

davidcheung (approved):

PR #164, opened 2020-11-25
add dev environment - initial commit

sshi100:
Major changes:
- add new policy 'k8s_developer_env_access' and assign to 'developer', Staging only
- add 'create-application-dev-user' in Makefile with a new secret template, that collects new users in IAM developer group, and create db, user and secret for each user

How to run:
- zero apply

Related changes in backend service repo
- added 'start-dev-env.sh' and skaffold.yaml


Makefile
```diff
@@ -19,8 +22,22 @@ create-application-user:
 	CREATE_SECRET=secret-application.yml.tpl \
 	sh ./db-ops/create-db-user.sh || echo
 
+create-application-dev-user:
```
bmonkman, line 25: This shouldn't use parameters passed in from `zero apply` like `randomSeed`, etc.
It's fine for the other ones since they only have to be run once, but you would need to need to run this make target every time you add a new dev to the team.

bmonkman: There should also be logic to not attempt this operation for users that already exist.

Makefile
```diff
@@ -19,8 +22,22 @@ create-application-user:
...
+	for u in ${CREATE_DB_DEV_USERS}; do \
+	REGION=${region} \
+	SEED=${randomSeed} \
+	PROJECT_NAME=${PROJECT_NAME} \
+	ENVIRONMENT=${ENVIRONMENT} \
+	NAMESPACE=${PROJECT_NAME} \
```
bmonkman, line 31: We probably shouldn't be creating these secrets in the real staging application namespace. How about somewhere else?
Do we even need these to be secrets? I'm assuming you're making secrets so you can copy them into the dev's namespace later, but how about just using aws secrets manager? If they are all in a namespace they will be lost if something happens to that namespace.

db-ops/secret-application-dev.yml.tpl
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: $USER_NAME
```
bmonkman, line 5: This template is exactly like the other one, just with a different variable here. How about instead of duplicating it we just change the variable we pass in? (Or remove this and use secrets manager as mentioned above)

templates/terraform/environments/stage/user_access.tf
```diff
@@ -173,6 +173,32 @@ data "aws_iam_policy_document" "deployer_access" {
...
+      resources = ["namespaces", "deployments", "deployments/scale", "configmaps", "pods", "pods/log", "pods/status", "pods/portforward", "pods/exec",
+        "jobs", "cronjobs",
+        "ingresses", "services", "serviceaccounts",
+        "replicasets", "horizontalpodautoscalers", "horizontalpodautoscalers/status"
+      ]
+    },
+    {
+      verbs      = ["get", "list", "watch"]
+      api_groups = ["*"]
+      resources = [ "daemonsets", "endpoints", "events",
```
bmonkman, line 191: Why just get,list,watch on daemonsets and pdbs? Seems like they should have the verbs listed above.

templates/terraform/environments/stage/user_access.tf
```diff
@@ -173,6 +173,32 @@ data "aws_iam_policy_document" "deployer_access" {
...
+        "poddisruptionbudgets"
+      ]
+    },
+    {
+      verbs      = ["create", "get", "list"]
```
bmonkman, line 196: Any reason not to have update, delete?

sshi100: Need to add 'kustomize' into installation check

templates/terraform/environments/stage/user_access.tf
sshi100: will provide more permissions to developers considering that Zero is designed for small team now and Staging can be taken as kind of Dev cloud environment. Eventually, we will do sub-namespace and allow full permission on sub-namespace only.

Makefile
sshi100: decided to use 1 user/pass for different user databases, then about the logic to check which user db has been created.

(from Bill) Sounds like we need to tweak the script in db-ops for this anyway, so maybe that script can check. Plus maybe we could just call it once with a list of all users. then inside the pod it loops through the users, checks for the db, creates any that don’t exist.

Makefile
sshi100: see above comment

db-ops/secret-application-dev.yml.tpl
sshi100: will use AWS secret manager instead of this one.

templates/terraform/environments/stage/user_access.tf
sshi100: see above reply.

Makefile
sshi100: if we create only one AWS user/pass secret for multiple databases... it's hard for zero apply Makefile to get a logic to know which databases inside RDS have been created.... as we don't suppose to access RDS and do show databases from laptop when zero apply.

sshi100: (fron Bill) Sounds like we need to tweak the script in db-ops for this anyway, so maybe that script can check. Plus maybe we could just call it once with a list of all users. then inside the pod it loops through the users, checks for the db, creates any that don’t exist

db-ops/job-create-db-postgres.yml.tpl
```diff
@@ -13,17 +13,10 @@ type: Opaque
...
-    SELECT 'DROP USER $DB_APP_USERNAME;' WHERE EXISTS (SELECT FROM pg_user WHERE usename = '$DB_APP_USERNAME');\gexec
-    CREATE USER $DB_APP_USERNAME WITH ENCRYPTED PASSWORD '$DB_APP_PASSWORD';
-    GRANT ALL PRIVILEGES ON DATABASE $DB_NAME TO $DB_APP_USERNAME;
-
----
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: $NAMESPACE
+    SELECT 'CREATE DATABASE __DB_NAME' WHERE NOT EXISTS (SELECT FROM pg_database WHERE datname = '__DB_NAME');\gexec
```
bmonkman, line 16: What's with the __ here?

db-ops/create-db-user.sh
```diff
@@ -53,6 +53,7 @@ fi
 JOB_ID=$(LC_ALL=C tr -dc 'a-z0-9' < /dev/urandom | head -c 8)
 eval "echo \"$(cat ./db-ops/job-create-db-${DATABASE_TYPE}.yml.tpl)\"" > ./k8s-job-create-db.yml
 [[ -z "${CREATE_SECRET}" ]] || eval "echo \"$(cat ./db-ops/${CREATE_SECRET})\"" >> ./k8s-job-create-db.yml
+[[ -z "${CREATE_DB_POD}" ]] || eval "echo \"$(cat ./db-ops/deployment-db-test.yml.tpl)\"" >> ./k8s-job-create-db.yml
```
bmonkman, line 56: Why was this pulled out into another file?

db-ops/job-create-db-postgres.yml.tpl
```diff
@@ -39,9 +32,14 @@ spec:
...
         - sh
-        args:
-        - '-c'
-        - psql -U$MASTER_RDS_USERNAME -h $DB_ENDPOINT postgres -v ON_ERROR_STOP=1 -a -f/db-ops/create-user.sql > /dev/null
+        - -c
+        - |
+          for db in $DB_LIST; do
+            [[ echo '\l' | psql -U$MASTER_RDS_USERNAME -h $DB_ENDPOINT postgres | grep \$db ]] || \
+            cat /db-ops/create-user.sql | \
+            sed \"s/__DB_NAME/\$db/g\" | \
```
bmonkman, line 40: Ah I see, you're using __ as a placeholder. How about we just stick with the same behaviour as in `create-db-user` and do proper variable substitution using `eval "echo \"$(cat /db-ops/create-user.sql)\"`?

templates/terraform/environments/stage/main.tf
```diff
@@ -41,6 +41,17 @@ data "terraform_remote_state" "shared" {
   }
 }
 
+# rds shared db password for dev envrionment
+module "rds_dev_secret" {
```
bmonkman, line 45: Honestly it would probably be fine to just set this password to be the project name or something, but if it's in a secret could you make sure that the dev group can read this secret?

db-ops/job-create-db-mysql.yml.tpl
```diff
@@ -13,16 +13,8 @@ type: Opaque
...
-    CREATE USER '$DB_APP_USERNAME' IDENTIFIED BY '$DB_APP_PASSWORD';
-    GRANT ALL PRIVILEGES ON $DB_NAME.* TO '$DB_APP_USERNAME';
-
----
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: $NAMESPACE
+    CREATE DATABASE IF NOT EXISTS __DB_NAME;
+    GRANT ALL PRIVILEGES ON __DB_NAME.* TO '$DB_APP_USERNAME' IDENTIFIED BY '$DB_APP_PASSWORD';
```
bmonkman, line 17: 👍

db-ops/job-create-db-postgres.yml.tpl
sshi100: it's a template format with __, modified to {{ }} for better understanding.

Makefile
sshi100: We have to suppose use `zero apply` everytime after add new dev, unless we move it to templates/ side but  db-ops/ is not there.

bmonkman: `db-ops` was moved into `templates/` in #163

db-ops/create-db-user.sh
sshi100: considered not to run db-pod for developers, so made it separate.  but now refreshed and removed.

db-ops/job-create-db-postgres.yml.tpl
sshi100: that __DB_NAME (now {{ DB_NAME }} can not be decided before the loop in Job.

templates/terraform/environments/stage/main.tf
sshi100: yes, dev group can read it

Makefile
sshi100: done the merge with templates/ one and put some enhancements for full review again.

templates/scripts/create-db-user.sh
```diff
@@ -6,41 +6,46 @@ usage () {
   exit 1
 }
 
+[[ "$1" == "devenv" ]] && DEVENV=$1
```
bmonkman, line 9: This is too obscure. Let's just accept SECRET_NAME and use that, and rely on the caller to set it.

templates/scripts/db-ops/job-create-db-postgres.yml.tpl
```diff
@@ -39,9 +38,14 @@ spec:
...
         - sh
-        args:
-        - '-c'
-        - psql -U$MASTER_RDS_USERNAME -h $DB_ENDPOINT postgres -v ON_ERROR_STOP=1 -a -f/db-ops/create-user.sql > /dev/null
+        - -c
+        - |
+          for db in $DB_NAME_LIST; do
+            [[ echo '\l' | psql -U$MASTER_RDS_USERNAME -h $DB_ENDPOINT postgres | grep \$db ]] || \
+            cat /db-ops/create-user.sql | \
+            sed \"s/{{ DB_NAME }}/\$db/g\" | \
```
bmonkman, line 46: mysql one is called VAR_DB, postgres is DB_NAME

sshi100: fixed

templates/scripts/create-dev-env.sh
```diff
@@ -0,0 +1,30 @@
...
+
+PROJECT=<% .Name %>
+AWS_DEFAULT_REGION=<% index .Params `region` %>
+RANDOM_SEED="<% index .Params `randomSeed` %>"
+ENVIRONMENT=stage  # only apply to Staging environment
+
+DEV_DB_LIST=$(aws iam get-group --group-name ${PROJECT}-developer-${ENVIRONMENT} | jq -r '"dev" + .Users[].UserName' | tr '\n' ' ')
+if [[ -z "${DEV_DB_LIST}" ]]; then
+    echo "$0: No developers available"
+    exit 2
```
bmonkman, line 11: On initial `zero apply` there won't be any users yet, and this non-zero exit will cause apply to fail.

templates/scripts/create-dev-env.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
```
bmonkman, line 1: This script should be chmodded to be executable.

templates/scripts/create-dev-env.sh
sshi100: yeah, let's exit with 0

bmonkman (approved):

PR #165, opened 2020-11-27
Add mail support for kratos using sendgrid

bmonkman:


templates/kubernetes/terraform/modules/kubernetes/user_auth.tf
```diff
@@ -96,6 +96,12 @@ resource "helm_release" "kratos" {
   #   # value = "${local.db_type}://${kubernetes_service.app_db.metadata[0].name}.${kubernetes_service.app_db.metadata[0].namespace}"
   # }
 
+  # This domain or address must be verified by the mail provider (Sendgrid, SES, etc.)
+  set {
+    name  = "kratos.config.courier.smtp.from_address"
+    value = "noreply@${var.domain_name}"
```
davidcheung, line 102: Just realized how hidden the `mail.` prefix is,
hmm maybe we need to have an output from the `sendgrid module`, or maybe remove the prefix all together from sendgrid side https://github.com/commitdev/terraform-aws-zero/blob/d0f855816a7f135ea02425982b6ca2b1d6a1f622/modules/sendgrid/main.tf#L34

templates/scripts/pre-k8s.sh
```diff
@@ -33,16 +33,21 @@ fi
...
+    SEED=${RANDOM_SEED} \
+    PROJECT_NAME=${PROJECT} \
+    ENVIRONMENT=${ENVIRONMENT} \
+    NAMESPACE=user-auth \
+    DATABASE_TYPE=<% index .Params `database` %> \
+    DATABASE_NAME=user_auth \
+    USER_NAME=kratos \
+    CREATE_SECRET=secret-user-auth.yml.tpl \
+    <%- if ne (index .Params `sendgridApiKey`) "" %>
+    SMTP_URI=smtps://apikey:${sendgridApiKey}@smtp.sendgrid.net:465 \
```
davidcheung, line 46: does this directly maps to kratos? 
in their docs it says [`COURIER_SMTP_CONNECTION_URI`](https://www.ory.sh/kratos/docs/reference/configuration)

bmonkman: No, it's used in `secret-user-auth.yml.tpl `

bmonkman: Yeah I don't think that prefix is useful, or maybe we leave it for flexibility but default it to `""`

templates/kubernetes/terraform/modules/kubernetes/user_auth.tf
bmonkman: Yeah I don't think that prefix is useful, or maybe we leave it for flexibility but default it to `""`

davidcheung: https://github.com/commitdev/terraform-aws-zero/pull/28

templates/scripts/pre-k8s.sh
davidcheung: Ah i see, I assume kratos recognizes `smtpConnectionURI`? couldnt find it in their docs

davidcheung (approved):

templates/kubernetes/terraform/modules/kubernetes/user_auth.tf
davidcheung: released v0.1.13

templates/kubernetes/terraform/modules/kubernetes/user_auth.tf
```diff
@@ -96,6 +96,11 @@ resource "helm_release" "kratos" {
   #   # value = "${local.db_type}://${kubernetes_service.app_db.metadata[0].name}.${kubernetes_service.app_db.metadata[0].namespace}"
   # }
 
+  set {
+    name  = "kratos.config.courier.smtp.from_address"
+    value = var.user_auth_mail_from_address
```
bmonkman, line 101: Changed this to be passed in per env in case they want to change it.

templates/scripts/pre-k8s.sh
bmonkman: The secret we are creating matches the one created by the helm chart, though we are doing it manually rather than through the chart: https://github.com/ory/k8s/blob/master/helm/charts/kratos/templates/secrets.yaml

davidcheung: ahh i see 👍  thanks for the explanation

PR #169, opened 2021-01-28
add NAT instance

sshi100:
add new NAT flag to with VPC creation

### DIFF
--- a/templates/terraform/environments/stage/main.tf
+++ b/templates/terraform/environments/stage/main.tf
@@ -102,7 +102,10 @@ module "stage" {
 
   domain_name = local.domain_name
 
-  # This will save some money as there a cost associated to each NAT gateway, but if the AZ with the gateway
+  # Instead of NAT gateway, enable NAT instance at cost of $1 per month
+  vpc_enable_nat_gateway = false
+
+  # When use NAT gateway, this setting will save some money as there a cost associated to each NAT gateway, but if the AZ with the gateway
   # goes down, nothing in the private subnets will be able to reach the internet. Not recommended for production.
   vpc_use_single_nat_gateway = true
 
--- a/templates/terraform/environments/stage/main.tf
+++ b/templates/terraform/environments/stage/main.tf
@@ -110,7 +110,7 @@ module "stage" {
   ## Use this setting to override default NAT instance types
   vpc_nat_instance_types = ["t3.nano", "t3a.nano"]
 
-  ## When use NAT gateway, this setting will save some money as there a cost associated to each NAT gateway, but if the AZ with the gateway goes down, nothing in the private subnets will be able to reach the internet. Not recommended for production.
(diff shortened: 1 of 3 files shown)


templates/terraform/environments/stage/main.tf
```diff
@@ -102,11 +102,14 @@ module "stage" {
...
 
-  # Instead of NAT gateway, enable NAT instance at cost of $1 per month
+  # NAT configuration
+  ## Instead of NAT gateway, enable NAT instance at cost of $1 per month
   vpc_enable_nat_gateway = false
 
-  # When use NAT gateway, this setting will save some money as there a cost associated to each NAT gateway, but if the AZ with the gateway
-  # goes down, nothing in the private subnets will be able to reach the internet. Not recommended for production.
+  ## Use this setting to override default NAT instance types
+  vpc_nat_instance_types = ["t3.nano", "t3a.nano"]
```
bmonkman, line 110: Let's just remove this and use the default. It won't be important for most people.

templates/terraform/environments/stage/main.tf
```diff
@@ -102,11 +102,14 @@ module "stage" {
 
   domain_name = local.domain_name
 
-  # Instead of NAT gateway, enable NAT instance at cost of $1 per month
+  # NAT configuration
+  ## Instead of NAT gateway, enable NAT instance at cost of $1 per month
```
bmonkman, line 106:
```suggestion
  # NAT configuration - NAT allows traffic from private subnets to access the public internet
  
  ## Instead of using AWS NAT gateway, use a NAT instance which is cheaper by about $30/month, though NAT gateways are more reliable. Only recommended for non-production environments.
```

templates/terraform/environments/stage/main.tf
```diff
@@ -102,11 +102,14 @@ module "stage" {
...
   vpc_enable_nat_gateway = false
 
-  # When use NAT gateway, this setting will save some money as there a cost associated to each NAT gateway, but if the AZ with the gateway
-  # goes down, nothing in the private subnets will be able to reach the internet. Not recommended for production.
+  ## Use this setting to override default NAT instance types
+  vpc_nat_instance_types = ["t3.nano", "t3a.nano"]
+
+  ## When use NAT gateway, this setting will save some money as there a cost associated to each NAT gateway, but if the AZ with the gateway goes down, nothing in the private subnets will be able to reach the internet. Not recommended for production.
```
bmonkman, line 112:
```suggestion
  ## When using NAT gateway, setting this to true will save some money as there a cost per-gateway of about $35/month. However, if the AZ with the gateway goes down nothing in the private subnets will be able to reach the internet. Not recommended for production. Not used if `vpc_enable_nat_gateway` is `false`.
```

templates/terraform/modules/environment/variables.tf
```diff
@@ -80,6 +80,12 @@ variable "vpc_use_single_nat_gateway" {
   default     = true
 }
 
+variable "vpc_nat_instance_types" {
+   description = "Candidates of instance type for the NAT instance"
+   type        = list
+   default     = ["t3.nano", "t3a.nano"]
```
bmonkman, line 86: Let's just provide t3.nano here. It's a bit confusing having 2 types here since we're not using spot instances.

bmonkman (approved):

Issue #170, opened 2021-01-29
Add notification service to eks stack


bmonkman:
Use helm to install the notification service if a sendgrid api key was provided

PR #171, opened 2021-01-29
Add notification service to eks stack (#170)

bmonkman:
Added sendgrid api key during setup
Add helm support for notification service
Some small formatting / text changes


templates/Makefile
```diff
@@ -77,8 +77,9 @@ teardown-secrets:
...
-	aws secretsmanager list-secrets --region $(AWS_DEFAULT_REGION) --query "SecretList[?Tags[?Key=='cf_keypair' && Value=='$(PROJECT)']].[Name] | [0][0]" | xargs aws secretsmanager delete-secret --region $(AWS_DEFAULT_REGION) --secret-id || echo "Secret already removed"
+	aws secretsmanager list-secrets --region $(AWS_DEFAULT_REGION) --query "SecretList[?Tags[?Key=='cf-keypair' && Value=='$(PROJECT)']].[Name] | [0][0]" | xargs aws secretsmanager delete-secret --region $(AWS_DEFAULT_REGION) --secret-id || echo "Secret already removed"
 	aws secretsmanager list-secrets --region $(AWS_DEFAULT_REGION) --query "SecretList[?Tags[?Key=='wg' && Value=='$(PROJECT)-$(ENVIRONMENT)']].[Name] | [0][0]" | xargs aws secretsmanager delete-secret --region $(AWS_DEFAULT_REGION) --secret-id || echo "Secret already removed"
+	aws secretsmanager list-secrets --region $(AWS_DEFAULT_REGION) --query "SecretList[?Tags[?Key=='sendgrid-api-key' && Value=='$(PROJECT)-$(ENVIRONMENT)']].[Name] | [0][0]" | xargs aws secretsmanager delete-secret --region $(AWS_DEFAULT_REGION) --secret-id || echo "Secret already removed"
```
davidcheung, line 82: line 79's should be deleting it by the `tag` already https://github.com/commitdev/zero-aws-eks-stack/blob/main/templates/terraform/bootstrap/secrets/main.tf#L44

templates/kubernetes/terraform/environments/stage/main.tf
```diff
@@ -80,4 +80,6 @@ module "kubernetes" {
   frontend_service_domain     = "<% index .Params `stagingFrontendSubdomain` %>${local.domain_name}"
   # This domain or address must be verified by the mail provider (Sendgrid, SES, etc.)
   user_auth_mail_from_address = "noreply@${local.domain_name}"
+
+  use_notification_service    = <%if eq (index .Params `sendgridApiKey`) "" %>false<% else %>true<% end %>
```
davidcheung, line 84: maybe `notification_service_enabled` ? 

we have a bunch of flags like 
`cf_signing_enabled ` (also declared in local as `file_uploads`)  
`sendgrid_enabled`
`auth_enabled`

templates/scripts/pre-k8s.sh
```diff
@@ -19,6 +20,18 @@ if [[ $? -ne 0 ]]; then
     fi
 fi
 
+<% if ne (index .Params `sendgridApiKey`) "" %>
```
davidcheung, line 23: oh i just realized this part, we also created it in https://github.com/commitdev/zero-aws-eks-stack/blob/main/templates/terraform/bootstrap/secrets/main.tf#L36-L45

templates/Makefile
bmonkman: Ohh I forgot that we are creating it there.. I'll remove the creation and deletion code I added, and use the one you already made.

templates/scripts/pre-k8s.sh
bmonkman: Yeah, oops. I'll pull this out.

templates/Makefile
bmonkman: Hmm the one in bootstrap will only be run once, so it expects that the api key will be reused between environments which might not be the case. I don't think it's a big issue right now though so I'll just stick with doing it that way.

davidcheung: yeah, didn't want to ask for 2 sendgrid api upfront, then just made one shared across envs back then, possibly need to revisit

davidcheung (approved):

sshi100 (approved):
Good too!

PR #182, opened 2021-02-24
Mode auth implementation into a module

davidcheung:
this enables someone to spin up extra auth instances and steps towards decoupling the infra / application logic more


templates/kubernetes/terraform/environments/prod/main.tf
```diff
@@ -75,13 +75,22 @@ module "kubernetes" {
...
 
   domain_name                 = local.domain_name
-  auth_enabled                = <% if eq (index .Params `userAuth`) "yes" %>true<% else %>false<% end %>
-  auth_domain                 = "auth.${local.domain_name}"
-  backend_service_domain      = "<% index .Params `productionBackendSubdomain` %>${local.domain_name}"
-  frontend_service_domain     = "<% index .Params `productionFrontendSubdomain` %>${local.domain_name}"
-  # This domain or address must be verified by the mail provider (Sendgrid, SES, etc.)
-  user_auth_mail_from_address = "noreply@${local.domain_name}"
-
+  user_auth = <% if eq (index .Params `userAuth`) "yes" %>[
```
bmonkman, line 78: How about just changing the variable to default to `[]` so we don't have to have the empty list here?

templates/kubernetes/terraform/modules/kubernetes/user_auth.tf
```diff
@@ -5,232 +5,18 @@ locals {
...
-    value = "admin.${var.auth_domain}"
-  }
-
-  set {
-    name  = "ingress.admin.tls[0].hosts[0]"
-    value = "admin.${var.auth_domain}"
-  }
+module "user_auth" {
+  count = length(var.user_auth)
+  source                      = "./user_auth"
```
bmonkman, line 12: Why not move this out to the modules repo?

bmonkman:
Looks like the `zero create` check failed. And when you figure out why we should see if we can add a guard around that case to give a better error message than a nil ptr dereference.

davidcheung:
Yeah I wrote a zero of `terraform-aws-zero` that is not yet released yet, i wonder if thats the problem 
not sure why it would fail at `zero create`, will dig deeper

templates/kubernetes/terraform/environments/stage/main.tf
```diff
@@ -74,13 +74,22 @@ module "kubernetes" {
...
+      backend_service_domain        = "<% index .Params `stagingBackendSubdomain` %>.${local.domain_name}"
+      jwks_secret_name              = "${local.project}-${local.environment}-oathkeeper-jwks-${local.random_seed}"
+      # This domain or address must be verified by the mail provider (Sendgrid, SES, etc.)
+      user_auth_mail_from_address   = "noreply@${local.domain_name}"
+    }
+    ## User auth: Kratos requires database and a secret (as: `user_auth[0].name`)
+    ## Oathkeeper requires a private key (as `user_auth[0].jwks_secret_name`)
+    ## per environment one of each (database/database secret/private key) is created in the pre-k8s step
+    ## If you need to add another user-auth instance you will have to create another set of these resources
+  ]<% else %>
```
davidcheung, line 92: oh it is probably this

bmonkman (approved):

bmonkman (approved):

PR #187, opened 2021-03-16
many fixes

davidcheung:
- removing empty deployment from db creation
- fix reference to undeclared variables
- allow notification secrets to partially exist
- fix user-auth domain name ended up with extra dot
- fix user-auth db-endpoint failure due to ext-svc


sshi100 (approved):
looked good to me

PR #206, opened 2021-06-28
Terraform 1.0 fixes and misc

bmonkman:


davidcheung (approved):
LGTM
assuming it should pass once terraform module is published

PR #208, opened 2021-07-06
Documentation Site

davidcheung:


doc-site/docusaurus.config.js
```diff
@@ -0,0 +1,112 @@
...
+
+// This is to download the CSS from commitdev/zero instead of using the local copy, this
+// allows us to maintain only one set of global css throughout the modules and have
+// consistent look and feel throughout the modules, this is synchronous but only runs during build time
+let customCss;
+try {
+  customCss = downloadCss(remoteCommonCssFile, cssDownloadPath)
+} catch (e) {
+  customCss = require.resolve('./src/css/fallback.css');
+}
```
davidcheung, line 20: The weird thing to note here is I'm downloading the global css file from `zero` repo then if it fails, it uses its original copy instead. Its a synchronous action but it only happens during build time in CI.

doc-site/sidebars.js
```diff
@@ -0,0 +1,58 @@
...
+        {
+          type: 'link',
+          label: 'Backend - Go',
+          href: `${config.url}/docs/modules/backend-go`,
+        },
+        {
+          type: 'link',
+          label: 'Backend - Node.js',
+          href: `${config.url}/docs/modules/backend-nodejs`,
+        }
```
davidcheung, line 54: I'm thinking of a way to consolidate all these repeated and should stay the same UI components, perhaps make a common npm package and use that.

bmonkman (approved):

bmonkman (approved):

PR #217, opened 2021-08-11
doc: user auth configuration docs and explanation

davidcheung:


doc-site/docs/components/kubernetes/user-auth.md
```diff
@@ -45,7 +45,46 @@ The default setup uses an Authenticator handler: [`cookie_session`][oathkeeper-c
 - Oathkeeper rules
 - jwks_private_key (Oathkeeper uses this key to sign session tokens)
 
-## Documentation
+## Configuring
+The default values can be overriden using the variable `kratos_values_override` and `oathkeeper_values_override`. You can pass in an object that is a subset of the Kratos or Oathkeeper config in the same nesting level and it will merge with the default values.
```
bmonkman, line 49: This is actually the helm chart config right? How about a link to the helm chart values so people can see all the possible options?

doc-site/docs/components/kubernetes/user-auth.md
```diff
@@ -45,7 +45,46 @@ The default setup uses an Authenticator handler: [`cookie_session`][oathkeeper-c
...
+
+| Name/Upstream | Routes | Purpose |
+| ---- | ----- | ------- |
+| Public Kratos | `/.ory/kratos/public` | Self serve auth flows to facilitate forms and redirects |
+| Admin Kratos | `/.ory/kratos/` | Handling request life cycle, only allows GET from external, other calls can be made internally in your cluster |
+| Backend public | `<(public\|webhook)\/.*>` | Public endpoints with no auth requirements |
+| Authenticated public | `<(?!(public\|webhook\|\.ory\/kratos)).*>` | Authenticated endpoints |
+
+## Oathkeeper
+View the possible configurations for [Kratos Configuration Reference](https://www.ory.sh/kratos/docs/v0.5/reference/configuration) and [Oathkeeper Configuration Reference](https://www.ory.sh/oathkeeper/docs/reference/configuration).
```
bmonkman, line 85:
```suggestion
For example if you want to change the Kratos error UI page you can override it as follows:
```hcl
kratos_values_override = {
  kratos = {
    config = {
      selfservice = {
        flows = {
          error = {
            ui_url = "https://<my-site.com>/custom-error-page"
          }
        }
      }
    }
  }
}
```

### Oathkeeper Proxy Rules
Oathkeeper rules are how you control auth decision making and routing through the proxy. Requests coming into the proxy only do something if they match a rule.
Each rule must have a **unique pattern matching string** (glob/regexp) and you can define which [handlers it must go through](https://www.ory.sh/oathkeeper/docs/pipeline) (Authenticators, Authorizers, Mutators, Error handlers), then at the end it can have an upstream service which is the destination of the requests (most likely your service).

:::caution
Incoming must match exactly 1 rule or Oathkeeper will throw an error.
:::

#### Zero's Proxy Rules setup
In our default setup there are 4 rules

| Name/Upstream | Routes | Purpose |
| ---- | ----- | ------- |
| Public Kratos | `/.ory/kratos/public` | Self serve auth flows to facilitate forms and redirects |
| Admin Kratos | `/.ory/kratos/` | Handling request life cycle, only allows GET from external, other calls can be made internally in your cluster |
| Backend public | `<(public\|webhook)\/.*>` | Public endpoints with no auth requirements |
| Authenticated public | `<(?!(public\|webhook\|\.ory\/kratos)).*>` | Authenticated endpoints |

## Oathkeeper
View the possible configurations for [Kratos Configuration Reference](https://www.ory.sh/kratos/docs/v0.5/reference/configuration) and [Oathkeeper Configuration Reference](https://www.ory.sh/oathkeeper/docs/reference/configuration).
```

doc-site/docs/components/kubernetes/user-auth.md
davidcheung: what helm chart doesnt really show the values, 
 https://github.com/ory/k8s/blob/master/helm/charts/kratos/values.yaml#L89 
i linked the https://github.com/commitdev/zero-aws-eks-stack/pull/217/files#diff-25dd24bdd30f2b5ade0d10f6d4d147456f340f1e3693f9804d16980a61397bdfR85 instead where its better documented what are the possible values, the entire thing just get mounted as a volume `/etc/config/kratos.yaml` i believe

bmonkman (approved):

PR #223, opened 2021-08-23
Notification svc external secret

davidcheung:


templates/terraform/bootstrap/secrets/main.tf
```diff
@@ -44,13 +44,34 @@ module "sendgrid_api_key" {
...
 
-  name  = "${local.project}-slack-<% index .Params `randomSeed` %>"
-  type  = "string"
-  value = var.slack_api_key
-  tags  = { slack: local.project }
+  name   = "${local.project}/kubernetes/prod/notification-service"
+  type   = "map"
+  values = {
+    SENDGRID_API_KEY  = var.sendgrid_api_key
+    SLACK_API_KEY     = var.slack_api_key,
```
bmonkman, line 56: single comma

templates/terraform/bootstrap/secrets/main.tf
```diff
@@ -44,13 +44,34 @@ module "sendgrid_api_key" {
...
+module "notification_service_secret_stage" {
+  count   = <%if eq (index .Params `notificationServiceEnabled`) "yes" %>1<% else %>0<% end %>
+  source  = "commitdev/zero/aws//modules/secret"
+  version = "0.0.2"
+
+  name   = "${local.project}/kubernetes/stage/notification-service"
+  type   = "map"
+  values = {
+    SENDGRID_API_KEY  = var.sendgrid_api_key
+    SLACK_API_KEY     = var.slack_api_key,
```
bmonkman, line 72: here too

templates/kubernetes/terraform/modules/kubernetes/notification_service.tf
```diff
@@ -62,14 +48,15 @@ resource "helm_release" "notification_service" {
...
 
-  set_sensitive {
-    name  = "application.slackApiKey"
-    value = var.notification_service_enabled && var.notification_service_slack_enabled ? data.aws_secretsmanager_secret_version.slack_api_key[0].secret_string : ""
+  set {
+    name  = "application.twilioPhoneNumber"
+    value = var.notification_service_twilio_phone_number
   }
+
+  depends_on = [data.aws_secretsmanager_secret.notification_service_secret]
```
bmonkman, line 61: I think this will be a syntax error because you have a count in the secret reference, so it will need to be `data.aws_secretsmanager_secret.notification_service_secret[0]`

bmonkman: Also this dependency doesn't really make sense. Typically you use `depends_on` with a resource because then you will wait until that resource exists. In the case of `data` a dependency check is not necessary because if the thing doesn't exist your terraform will fail at line 6 in this file.

bmonkman (changes requested):

templates/kubernetes/terraform/modules/kubernetes/notification_service.tf
davidcheung: because both have them have `count` based on `notification_service_enabled` so in both `enabled` and `disabled` they wont error out, but in the case someone has it enabled but don't have the secret then it would fail. 

I wanted to surface the error earlier so it doesn't need to fail at the deployment step, I guess having it fail there is okay, will remove.

bmonkman: You could also leave in the data reference but then remove the dependency but then it would still be failing at the terraform apply which has its pros and cons.. It's closer to the actual failure so potentially easier for us to give timely feedback, but erroring out and re-running the terraform isn't a great experience.

bmonkman (approved):

templates/kubernetes/terraform/modules/kubernetes/notification_service.tf
davidcheung: I think its fine removing, a fresh install shouldn't run into the missing secret at all, in the rare case that happens I guess they would be able to dig into why `helm deploy` has failed

PR #226, opened 2021-09-03
dev env: fix secret creation for external secret

davidcheung:
also defaults VPN IP to 201 when none exist
and update exec to new exec format (since `exec -it <pod> sh`) is deprecated

### DIFF
--- a/templates/scripts/add-vpn-user.sh
+++ b/templates/scripts/add-vpn-user.sh
@@ -7,6 +7,12 @@ CLUSTER=$(kubectl config current-context | cut -d"/" -f2)
 NAMESPACE=<% .Name %>
 REGION=<% index .Params `region` %>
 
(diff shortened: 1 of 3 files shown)


bmonkman (approved):

templates/scripts/add-vpn-user.sh
```diff
@@ -16,7 +22,7 @@ if [ -z "$POD" ]; then
 fi
 
 function k8s_exec() {
-  kubectl -n vpn exec -it $POD -- /bin/bash -c "$1"
+  kubectl -n vpn exec $POD wireguard --container wireguard -- /bin/bash -c "$1"
```
bmonkman, line 25: Does the wireguard pod have multiple containers in it?

davidcheung: yeah somehow it has a `sysctls` container in there

PR #227, opened 2021-09-08
provision dev-env user-auth for local development

davidcheung:


templates/kubernetes/terraform/environments/stage/main.tf
```diff
@@ -106,7 +106,9 @@ module "kubernetes" {
     ## Oathkeeper requires a private key (as `user_auth[0].jwks_secret_name`)
     ## per environment one of each (database/database secret/private key) is created in the pre-k8s step
     ## If you need to add another user-auth instance you will have to create another set of these resources
-  ]<% end %>
+  ]
+  enabled_dev_env_user_auth = true
```
bmonkman, line 110: maybe `user_auth_dev_env_enabled`? And could you add a comment?

templates/kubernetes/terraform/modules/kubernetes/user_auth.tf
```diff
@@ -36,3 +36,74 @@ module "user_auth" {
 
   depends_on = [helm_release.external_secrets]
 }
+
+module "dev_user_auth" {
```
bmonkman, line 40: There's not a conditional in here to use the flag to enable/disable this.

templates/kubernetes/terraform/modules/kubernetes/variables.tf
```diff
@@ -144,6 +144,24 @@ variable "user_auth" {
...
+}
+
+variable "dev_user_auth_frontend_domain" {
+  description = "Frontend domain used for local development with dev env"
+  type        = string
+  default     = "127.0.0.1:3000"
+}
+
+variable "dev_user_auth_frontend_scheme" {
+  description = "Frontend domain used for local development with dev env"
```
bmonkman, line 160: Wrong description

templates/kubernetes/terraform/modules/kubernetes/variables.tf
```diff
@@ -144,6 +144,24 @@ variable "user_auth" {
...
+  default     = false
+}
+
+variable "dev_user_auth_frontend_domain" {
+  description = "Frontend domain used for local development with dev env"
+  type        = string
+  default     = "127.0.0.1:3000"
+}
+
+variable "dev_user_auth_frontend_scheme" {
```
bmonkman, line 159: AFAICT this scheme and domain are only ever concatenated, never used independently. Is there a reason to keep them separate in the vars?

davidcheung: just checked, its always used together 
hmm I guess we should merge them, the only downside would be is its a breaking change

bmonkman (approved):

PR #250, opened 2022-01-17
Serverless initial implementation [MAJOR]

davidcheung:
Notable breaking changes
- requires new zero binary version for parameter option `elseValue` 
- database user creation is no longer dependent on k8s instead will use Lambda + docker image


bmonkman:
Looks like there's an issue with the tf validate check, is it legit?

bmonkman (approved):

davidcheung:
thanks! 
it was legit, the prod `user_access` had a syntax error

templates/scripts/pre-k8s.sh
```diff
@@ -11,7 +11,7 @@ OATHKEEPER_SECRET_NAME=${PROJECT}-${ENVIRONMENT}-oathkeeper-jwks-${RANDOM_SEED}
 aws secretsmanager describe-secret --region ${AWS_DEFAULT_REGION} --secret-id ${VPN_SECRET_NAME} > /dev/null 2>&1
 if [[ $? -ne 0 ]]; then
     echo "Creating VPN private key..."
-    secret=$(kubectl run --rm --quiet --attach=true --context ${KUBE_CONTEXT} zero-k8s-utilities --image=commitdev/zero-k8s-utilities:${ZERO_K8S_UTILS_VERSION} --restart=Never -- wg genkey)
+    secret=$(docker run commitdev/zero-k8s-utilities:${ZERO_K8S_UTILS_VERSION} "sh" "-c"  "wg genkey")
```
aspcanada, line 14: Did you update the prerequisites to include `docker`? Or is that assumed with `kubectl`?

Makefile
```diff
@@ -5,7 +5,7 @@ SHELL := /bin/bash
 run: make-apply
 
 make-apply:
-	cd $(PROJECT_DIR) && AUTO_APPROVE="-auto-approve" make
+	@export AUTO_APPROVE="-auto-approve" && cd $(PROJECT_DIR) && (if [[ "${backendApplicationHosting}" =~ "kubernetes" ]]; then make apply; else make apply-without-eks; fi)
```
aspcanada, line 8: Are you going to implement a `teardown` associated with SAM?

templates/terraform/modules/environment/lambda-db-ops/vairables.tf
```diff
@@ -0,0 +1,18 @@
...
+variable "subnet_ids" {
+  description = "Subnet IDs the lambda function is able to connect to inside the VPC"
+  default     = []
+  type        = list(string)
+}
+variable "security_group_ids" {
+  description = "Security groups the lambda function is able to connect to inside the VPC"
+  default     = []
+  type        = list(string)
+}
```
aspcanada, line 18: This filename has a typo

templates/terraform/modules/environment/lambda-db-ops/vairables.tf
```diff
@@ -0,0 +1,18 @@
+variable "project" {
+  description = "The name of the project, mostly for tagging"
+}
+
+variable "environment" {
+  description = "The environment (dev/staging/prod)"
```
aspcanada, line 6: this is the first time seeing a "dev" envir in Zero.

templates/terraform/modules/environment/main.tf
```diff
@@ -52,8 +41,9 @@ data "aws_caller_identity" "current" {}
 #
 # Provision the EKS cluster
 module "eks" {
+  count = var.serverless_enabled ? 0 : 1
   source  = "commitdev/zero/aws//modules/eks"
```
aspcanada, line 45: two forward slashes here... is that on purpose?

templates/terraform/modules/environment/sam/main.tf
```diff
@@ -0,0 +1,67 @@
...
+      rule {
+        apply_server_side_encryption_by_default {
+          sse_algorithm = "AES256"
+        }
+      }
+    }
+}
+
+module "ecr" {
+  source  = "commitdev/zero/aws//modules/ecr"
```
aspcanada, line 30: double //

templates/terraform/modules/environment/sam/variables.tf
```diff
@@ -0,0 +1,32 @@
+
+variable "project" {
+  description = "The name of the project, mostly for tagging"
+}
+
+variable "environment" {
+  description = "The environment (dev/staging/prod)"
```
aspcanada, line 7: dev env?

templates/scripts/pre-k8s.sh
bmonkman: Yeah we are going to have to update the dependencies in the zero binary repo.

templates/terraform/modules/environment/main.tf
bmonkman: Yes, this syntax is used to refer to a sub module. https://www.terraform.io/language/modules/sources#modules-in-package-sub-directories

Makefile
bmonkman: This is the makefile at the root of the module, the one in the templates dir is responsible for teardown. But yeah, we should probably have an equivalent `teradown-without-eks` make target that just doesn't run the teardown eks stuff.

templates/terraform/modules/environment/lambda-db-ops/vairables.tf
bmonkman: It won't affect anything because terraform just concatenates all the files in the directory regardless of their names but yeah, it should be fixed.

davidcheung:
rebased against main to resolve the conflict

PR #251, opened 2022-01-20
2 small fixes

pmoranga:
This PR fixes 2 issues when using AWS backend.

1. Shell interpreter has issue when sourcing file.
2. Add check to correlate with conditional Cloud Trail.

### DIFF
--- a/templates/terraform/environments/shared/main.tf
+++ b/templates/terraform/environments/shared/main.tf
@@ -174,5 +174,5 @@ output "ci_user_name" {
 }
 
 output "cloudtrail_bucket_id" {
-  value = local.enable_cloudtrail > 0 ? module.cloudtrail[0].cloudtrail_bucket_id : ""
+  value = local.enable_cloudtrail > 0 ? module.cloudtrail[0].cloudtrail_bucket_id : "not-applicable"
 }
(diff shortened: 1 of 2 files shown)


bmonkman (approved):
Thanks for your contribution!

templates/terraform/environments/shared/main.tf
```diff
@@ -174,5 +174,5 @@ output "ci_user_name" {
 }
 
 output "cloudtrail_bucket_id" {
-  value = module.cloudtrail.cloudtrail_bucket_id
+  value = local.enable_cloudtrail > 0 ? module.cloudtrail[0].cloudtrail_bucket_id : ""
```
bmonkman, line 177: This would fail later when this value is used since it will result in an invalid ARN in IAM. The output would need to be something non-empty. Maybe just `not-applicable` or something.

bmonkman (changes requested):

templates/terraform/environments/shared/main.tf
pmoranga: will it be used if `enable_cloudtrail` is disabled ? I used it with disable without issue.

bmonkman (approved):
Thanks!

PR #267, opened 2022-05-16
Fix issue with Secret not found, keep both permissions.

pmoranga:
… needed.


bmonkman (approved):
Yep, this should help with backward compatibility. Thanks for your contribution!

